PR #10, opened 2023-11-22
fix: properly override a default component

joelhooks:
this also required wrangling some types. thats @delucis for the clues!


delucis:
Yay! Also hoping to get an official API out for extending the docs schema end of this week or start of next. That will help clean up `src/content/config.ts` for cases like this when it's ready.

PR #38, opened 2023-12-19
feat: adds an image resource to sanity that gets created on cloudinary upload

joelhooks:


apps/course-builder-web/src/inngest/functions/cloudinary/cloudinary-webhooks-handlers.ts
```diff
@@ -11,6 +11,25 @@ export const cloudinaryAssetCreated = inngest.createFunction(
...
   async ({event, step}) => {
+    const imageResource = await step.run(
+      'create the video resource in Sanity',
+      async () => {
+        await sanityMutation([
+          {
+            createOrReplace: {
+              _id: event.data.asset_id,
+              _type: 'imageResource',
+              url: event.data.secure_url,
```
joelhooks, line 22: kept this simple, with the asset id as the sanity id we could do a lookup if wanted/needed to later

apps/course-builder-web/src/server/sanity.server.ts
```diff
@@ -37,20 +37,26 @@ export async function sanityMutation(
...
+    }.sanity.io/v${env.SANITY_STUDIO_API_VERSION}/data/query/${
+      env.SANITY_STUDIO_DATASET
+    }?query=${encodeURIComponent(query)}&perspective=published`,
     {
       method: 'get',
       headers: {
         Authorization: `Bearer ${env.SANITY_API_TOKEN}`,
       },
-      next: {revalidate: 60}, //seconds
+      next: {revalidate: options.revalidate}, //seconds
```
joelhooks, line 59: this makes things weird if it's not set low

PR #41, opened 2023-12-20
fix: fallback user.role should be 'user' string

jbranchaud:
It was instead being assigned the `user` object that was one of the
arguments to the `session` callback function. I'm unsure why this wasn't
resulting in a type error.

role

### DIFF
--- a/apps/course-builder-web/src/server/auth.ts
+++ b/apps/course-builder-web/src/server/auth.ts
@@ -12,6 +12,8 @@ import {mysqlTable} from '@/server/db/schema'
 import {inngest} from '@/inngest/inngest.server'
 import {USER_CREATED_EVENT} from '@/inngest/events'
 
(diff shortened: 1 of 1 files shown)


apps/course-builder-web/src/server/auth.ts
```diff
@@ -48,12 +50,15 @@ export const authOptions: NextAuthOptions = {
...
       })
+
+      const role: Role = dbUser?.role || 'user'
+
       return {
         ...session,
         user: {
           ...session.user,
           id: user.id,
-          role: dbUser?.role || user,
```
jbranchaud, line 56: I don't understand why this wasn't resulting in a type error somewhere for the session callback since it seems to have a known type and the `UserAdapter` type is not part of the union for `user.role`.

joelhooks (approved):

PR #43, opened 2023-12-21
feat: redirect unauthorized tip CRUD to /tips

jbranchaud:
This does a server-redirect to the `/tips` index page if you are unauthorized the do CRUD on tips. Instead of seeing a dead-end blank page, you land back on a page with content.

With the Page Router we would typically do this sort of auth-check and redirect in `getServerSideProps`

```javascript
export async function getServerSideProps(context) {
  const session = await getServerAuthSession()
  const ability = getAbility({user: session?.user})

  if (!ability.can('upload', 'Media')) {
    return {
      redirect: {
        destination: '/',
        permanent: false,
      },
    }
  }
 
  return {
    props: {},
  }
}
```

What we can now do with a `redirect` call in a server-side page/component function of the App Router is a nice ergonomics improvement.

land


joelhooks (approved):

apps/course-builder-web/src/lib/ability.ts
```diff
@@ -16,15 +15,10 @@ export const UserSchema = z.object({
...
 
-type Abilities =
-  | ['view', 'Anything']
-  | ['view', 'Profile' | User]
-  | ['upload', 'Media']
-  | ['edit', 'Tip']
-  | ['edit', 'Article']
-  | ['edit', 'Module']
+type Actions = 'create' | 'read' | 'update' | 'delete' | 'manage'
+type Subjects = 'Content' | 'User' | User | 'all'
```
joelhooks, line 19: had "refine these" on my todo list. Probably want to go back through CASL docs since we are using older style for some stuff.

PR #44, opened 2023-12-21
refactor: try groqd for typed sanity queries

jbranchaud:
I thought this would be an interesting spike. I'm not sure if I'm sold on it yet or not.

Pros:
- Zod parsing/types are baked into the groqd API, so that stays in sync as you evolve the query
- easier to share/reuse partial queries

Cons:
- you kinda have to do two layers of mental translation: `[data I want] --> groq --> groqd`
- groqd doesn't support the full groq syntax, not clear on where those edges are yet

groq

### DIFF
--- a/apps/course-builder-web/package.json
+++ b/apps/course-builder-web/package.json
@@ -100,6 +100,7 @@
     "drizzle-orm": "^0.28.5",
     "framer-motion": "^10.16.5",
     "groq": "^3.18.1",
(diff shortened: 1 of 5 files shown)


jbranchaud:
`groqd` was worth trying out, but I think the limitations to the query language are going to conflict with some of our query patterns. That will make adoption tricky and having two ways to write queries scattered around the codebase seems more like tech debt than an enhancement.

PR #51, opened 2024-01-13
Add `.env.example`

chaance:
The docs mention copying this file but it doesn't exist. So now it does!


joelhooks:
it's over here, the top-level readme needs some love https://github.com/badass-courses/course-builder/blob/main/apps/course-builder-web/.env.example

PR #57, opened 2024-01-31
feat(articles): add state and visibility fields

vojtaholik:
image

gif


apps/course-builder-web/src/app/articles/_components/edit-article-form.tsx
```diff
@@ -131,9 +148,58 @@ export function EditArticleForm({article}: {article: Article}) {
...
+                      </FormControl>
+                      <SelectContent className="">
+                        {ArticleStateSchema.options.map((option) => {
+                          const value = option._def.value
+                          return (
+                            <SelectItem key={value} value={value}>
+                              {value}
+                            </SelectItem>
+                          )
+                        })}
```
vojtaholik, line 182: schema driven ™

apps/course-builder-web/src/trpc/api/routers/articles.ts
```diff
@@ -86,27 +87,21 @@ export const articlesRouter = createTRPCRouter({
...
     }),
   update: protectedProcedure
-    .input(
-      z.object({
-        articleId: z.string(),
-        title: z.string(),
-        body: z.string().optional().nullable(),
-      }),
-    )
+    .input(ArticleFormSchema)
```
vojtaholik, line 90: this can just take `ArticleFormSchema` to consolidate the types

joelhooks: that's probably an indicator that `ArticleFormSchema` could be promoted to `ArticleSchema`

joelhooks: actually that already exists and should be used: https://github.com/badass-courses/course-builder/blob/57/apps/course-builder-web/src/lib/articles.ts#L17-L27

joelhooks: ahh I see because we have extra validations

joelhooks (approved):

PR #59, opened 2024-01-31
feat: partial pre-rendering

joelhooks:
this explores using [nextjs partial pre-rendering](https://nextjs.org/learn/dashboard-app/partial-prerendering)


apps/course-builder-web/src/app/[article]/page.tsx
```diff
@@ -34,32 +35,57 @@ export async function generateMetadata({ params, searchParams }: Props, parent:
...
 
-export default async function ArticlePage({ params }: { params: { article: string } }) {
+async function ArticleActionBar({ slug }: { slug: string }) {
   const session = await getServerAuthSession()
   const ability = getAbility({ user: session?.user })
-  const article = await getArticle(params.article)
-
-  if (!article) {
-    notFound()
-  }
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/[article]/page.tsx:44]: I moved all the dynamic stuff down into components so that the page won't be blocked completely while loading. Not 100% sure this is the way, but it looks ok from how I understand it.

apps/course-builder-web/src/app/[article]/page.tsx
```diff
@@ -34,32 +35,57 @@ export async function generateMetadata({ params, searchParams }: Props, parent:
...
+
+  return <h1 className="text-3xl font-bold">{article?.title}</h1>
+}
+
+export default async function ArticlePage({ params }: { params: { article: string } }) {
+  return (
+    <div>
+      <Suspense fallback={<div className="bg-muted flex h-9 w-full items-center justify-between px-1" />}>
+        <ArticleActionBar slug={params.article} />
+      </Suspense>
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/[article]/page.tsx:85]: the action bar loads based on user data so I'm wrapping it with suspense which is the signal to next that this is a dynamic section and we don't need to wait for it

apps/course-builder-web/src/app/[article]/page.tsx
```diff
@@ -34,32 +35,57 @@ export async function generateMetadata({ params, searchParams }: Props, parent:
...
+      <ReactMarkdown>{article.description}</ReactMarkdown>
+    </div>
+  )
+}
+
+async function ArticleTitle({ slug }: { slug: string }) {
+  const article = await getArticle(slug)
+
+  return <h1 className="text-3xl font-bold">{article?.title}</h1>
+}
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/[article]/page.tsx:78]: > Move data fetching down to the components that need it, thus isolating which parts of your routes should be dynamic in preparation for Partial Prerendering.

From [the next docs](https://nextjs.org/learn/dashboard-app/partial-prerendering) these sub-components load the data that they need. They could probably be combined, BUT the article calls should be well cached

apps/course-builder-web/src/app/[article]/page.tsx
Comment (leerob) [reply apps/course-builder-web/src/app/[article]/page.tsx:44]: This looks right, since this component is looking at the incoming request (for auth) so it's dynamic.

apps/course-builder-web/src/app/[article]/page.tsx
```diff
@@ -34,32 +35,57 @@ export async function generateMetadata({ params, searchParams }: Props, parent:
...
+    </>
+  )
+}
 
-        <div className="mt-4 pb-32">
-          <ReactMarkdown className="prose dark:prose-invert">{article.body}</ReactMarkdown>
-          <ReactMarkdown>{article.description}</ReactMarkdown>
-        </div>
+async function Article({ slug }: { slug: string }) {
+  const article = await getArticle(slug)
```
Comment (leerob) [inline apps/course-builder-web/src/app/[article]/page.tsx:60]: And assuming this is using `fetch`, or `unstable_cache`, it can be cached and prerendered during the build.

Comment (joelhooks) [reply apps/course-builder-web/src/app/[article]/page.tsx:60]: @leerob thats where it gets confusing to me with the dynamic aspects and how my airhead abilities will play with content loading

here we are working on the Article because it's the simplest media format we deal with but we still have access and visibility controls

the rules get more complicated as we expand into the courses and commerce, but once i get the pattern straight we should be off to the races

vojtaholik (approved):

PR #68, opened 2024-02-12
feat: only create a single promise and pass it down as a loader

joelhooks:
this lets us cross the client barrier with the promise and use the `use` hook to stream the results in

less blocking overall!

### DIFF
--- a/apps/course-builder-web/src/app/tips/[slug]/page.tsx
+++ b/apps/course-builder-web/src/app/tips/[slug]/page.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react'
-import { Suspense } from 'react'
+import { Suspense, use } from 'react'
 import type { Metadata, ResolvingMetadata } from 'next'
 import Link from 'next/link'
 import { notFound } from 'next/navigation'
 import { TipPlayer } from '@/app/tips/_components/tip-player'
 import { getAbility } from '@/lib/ability'
-import { getTip } from '@/lib/tips'
+import { getTip, Tip } from '@/lib/tips'
 import { VideoResource } from '@/lib/video-resource'
 import { getServerAuthSession } from '@/server/auth'
 import { sanityQuery } from '@/server/sanity.server'
@@ -29,7 +29,7 @@ export async function generateMetadata({ params, searchParams }: Props, parent: 
 
   const previousImages = (await parent).openGraph?.images || []
 
-  const ogImage = getOGImageUrlForResource(tip)
+  const ogImage = getOGImageUrlForResource(tip, 'tips')
 
   return {
     title: tip.title,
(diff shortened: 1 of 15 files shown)


apps/course-builder-web/src/app/tips/[slug]/page.tsx
```diff
@@ -98,17 +99,17 @@ function PlayerContainerSkeleton() {
...
 
   if (!tip) {
     notFound()
   }
 
-  const videoResource = await sanityQuery<VideoResource>(
+  const videoResourceLoader = sanityQuery<VideoResource>(
     `*[_type == "videoResource" && _id == "${tip.videoResourceId}"][0]`,
     { tags: ['tips', tip._id] },
   )
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/tips/[slug]/page.tsx:114]: this gets passed to a _client_ component below, but instead of blocking with `await`, we can push the promise onto the client

apps/course-builder-web/src/app/tips/_components/tip-player.tsx
```diff
@@ -26,6 +27,8 @@ export function TipPlayer({
     minResolution: '540p',
   } as MuxPlayerProps
 
+  const videoResource = use(videoResourceLoader)
```
joelhooks, line 30: `use` is a hook that will resolve the promise for us that the server passed on.

this is voodoo magic as far as i am concerned, but the cool kind 🤡

it doesn't block the ui and streams in the result

apps/course-builder-web/src/app/tips/[slug]/page.tsx
```diff
@@ -40,20 +40,20 @@ export async function generateMetadata({ params, searchParams }: Props, parent:
...
       <main className="mx-auto w-full" id="tip">
         <Suspense fallback={<div className="bg-muted flex h-9 w-full items-center justify-between px-1" />}>
-          <TipActionBar slug={params.slug} />
-        </Suspense>
-        <Suspense fallback={<PlayerContainerSkeleton />}>
-          <PlayerContainer slug={params.slug} />
+          <TipActionBar slug={params.slug} tipLoader={tipLoader} />
         </Suspense>
+
+        <PlayerContainer slug={params.slug} tipLoader={tipLoader} />
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/tips/[slug]/page.tsx:51]: the `Suspense` boundary is actually a level lower

apps/course-builder-web/src/app/tips/[slug]/page.tsx
```diff
@@ -98,40 +98,42 @@ function PlayerContainerSkeleton() {
...
+                hidden: displayOverlay,
+              })}
+            >
+              <TipPlayer videoResourceLoader={videoResourceLoader} />
+            </div>
           </div>
         </div>
       </div>
-    </div>
+    </Suspense>
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/tips/[slug]/page.tsx:131]: suspending here because we don't depend on the `Tip` we depend on the `VideoResource`

PR #71, opened 2024-02-13
feat(workflow): generate transcript with screenshots from video

vojtaholik:
data preview screenshot
inngest function screenshot

capturing this for later gif


apps/course-builder-web/src/inngest/functions/ai/data/generate-transcript-with-screnshots.ts
```diff
@@ -0,0 +1,153 @@
...
+      throw new NonRetriableError(`Video resource not found for id (${event.data.videoResourceId})`)
+    }
+    const transcriptWithScreenshots = await step.run('generate transcript with screenshots', async () => {
+      return await mergeSrtWithScreenshots(videoResource.srt, videoResource.muxPlaybackId)
+    })
+    return { transcriptWithScreenshots }
+  },
+)
+
+async function mergeSrtWithScreenshots(srt: string, muxPlaybackId: string) {
```
vojtaholik, line 30: this is based on @tayiorbeii’s [screenshot-from-srt-args](https://github.com/skillrecordings/transcript-automation-kenv/blob/main/scripts/screenshot-srt-from-args.js) script

apps/course-builder-web/src/lib/resources.ts
```diff
@@ -9,5 +9,8 @@ export async function getResource<T = any>(slugOrId: string): Promise<T | null>
           summary,
           body,
           "slug": slug.current,
+          srt,
+          wordLevelSrt,
+          muxPlaybackId,
   }`)
```
joelhooks, line 15: something to keep in mind/recognize here is that `getResource` was **type agnostic** but this wholly makes this `getVideoResource` and should be typed and moved into it's own lib

apps/course-builder-web/src/trpc/api/routers/tips.ts
```diff
@@ -145,4 +145,24 @@ export const tipsRouter = createTRPCRouter({
...
+    )
+    .mutation(async ({ ctx, input }) => {
+      const session = await getServerAuthSession()
+      const ability = getAbility({ user: session?.user })
+      if (!ability.can('create', 'Content')) {
+        throw new Error('Unauthorized')
+      }
+
+      return await inngest.send({
+        name: TRANSCRIPT_WITH_SCREENSHOTS_REQUESTED_EVENT,
```
joelhooks, line 162: with Inngest, and more broadly with event driven workflows, you want to tell the system what has happened, and not request specific work

imo, this should simply occur with `VIDEO_UPLOADED` or `TRANSCRIPT_COMPLETE` and not as a remote procedure call (rpc via tRPC)

apps/course-builder-web/src/inngest/functions/ai/data/generate-transcript-with-screnshots.ts
joelhooks: this needs to be extracted into a `transcript-processing` folder that we could move other similar utils in to and create a package from

joelhooks (approved):

PR #75, opened 2024-02-16
feat: give me selectable prompts

joelhooks:
select

### DIFF
--- a/apps/course-builder-web/src/app/_components/resource-assistant.tsx
+++ b/apps/course-builder-web/src/app/_components/resource-assistant.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import { useRef } from 'react'
+import { AssistantWorkflowSelector } from '@/app/_components/assistant-workflow-selector'
 import { ResourceChatResponse } from '@/app/_components/resource-chat-response'
 import { useSocket } from '@/hooks/use-socket'
 import { EnterIcon } from '@sanity/icons'
@@ -10,14 +11,20 @@ import { Button, Textarea } from '@coursebuilder/ui'
 export function ResourceAssistant({
   resourceId,
   handleSendMessage,
+  availableWorkflows = [{ value: 'summarize', label: 'Summarize', default: true }],
 }: {
   resourceId: string
   handleSendMessage: (
     event: React.KeyboardEvent<HTMLTextAreaElement> | React.MouseEvent<HTMLButtonElement, MouseEvent>,
     messages: ChatCompletionRequestMessage[],
+    selectedWorkflow?: string,
   ) => void
+  availableWorkflows?: { value: string; label: string; default?: boolean }[]
 }) {
   const [messages, setMessages] = React.useState<ChatCompletionRequestMessage[]>([])
+  const [selectedWorkflow, setSelectedWorkflow] = React.useState<string>(
+    availableWorkflows.find((w) => w.default)?.value || 'summarize',
+  )
 
   useSocket({
     room: resourceId,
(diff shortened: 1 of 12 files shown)


apps/course-builder-web/src/app/_components/resource-assistant.tsx
```diff
@@ -10,14 +11,20 @@ import { Button, Textarea } from '@coursebuilder/ui'
 export function ResourceAssistant({
   resourceId,
   handleSendMessage,
+  availableWorkflows = [{ value: 'summarize', label: 'Summarize', default: true }],
```
joelhooks, line 14: `summarize` is the slug of a generic workflow/prompt

apps/course-builder-web/src/app/_components/resource-assistant.tsx
```diff
@@ -37,7 +44,18 @@ export function ResourceAssistant({
...
+        <h3 className="inline-flex p-5 pb-3 text-lg font-bold">Assistant</h3>
+        {availableWorkflows.length > 1 && (
+          <AssistantWorkflowSelector
+            initialValue={selectedWorkflow}
+            onValueChange={(value) => {
+              setSelectedWorkflow(value || selectedWorkflow)
+            }}
+            availableWorkflows={availableWorkflows}
+          />
+        )}
```
joelhooks, line 57: we only want this if it has more than 1 to choose from, otherwise it's hidden

apps/course-builder-web/src/inngest/helpers/resource-chat.ts
```diff
@@ -40,40 +40,74 @@ export async function resourceChat({
...
-        content: systemPromptAction.content,
-      }
+  let systemPrompt: ChatCompletionRequestMessage = {
+    role: systemPromptAction.role,
+    content: systemPromptAction.content,
+  }
+  let seedMessages: ChatCompletionRequestMessage[] = []
+
+  try {
+    const actionParsed = JSON.parse(systemPromptAction.content)
```
joelhooks: here's where it gets a little spicy 🌶️

taylor had a prompt sequence that simulated a little back and forth between the llm and the user to prime the convo so this parses the content of the prompt action. If it throws an error (not valid json mostly) it will drop down to the assumption that we just want the text of the `content`

apps/course-builder-web/src/inngest/helpers/resource-chat.ts
```diff
@@ -40,40 +41,88 @@ export async function resourceChat({
...
+          role: z.enum([
+            ChatCompletionRequestMessageRoleEnum.System,
+            ChatCompletionRequestMessageRoleEnum.User,
+            ChatCompletionRequestMessageRoleEnum.Assistant,
+            ChatCompletionRequestMessageRoleEnum.Function,
+          ]),
+          content: z.string(),
+        }),
+      )
+      .parse(JSON.parse(systemPromptAction.content))
```
joelhooks, line 73: here's where it gets a little spicy 🌶️

taylor had a prompt sequence that simulated a little back and forth between the llm and the user to prime the convo so this parses the content of the prompt action. If it throws an error (not valid json mostly) it will drop down to the assumption that we just want the text of the content

right now i am assuming any parsed json is an array of Chat Completion Request Messages and use zod to enforce that. Anthing else will get booted

this is fine because gpt-4 will take it as text too, this just gives us a clean array of messages

apps/course-builder-web/src/inngest/helpers/resource-chat.ts
```diff
@@ -40,40 +41,88 @@ export async function resourceChat({
...
+        return {
+          role: systemPromptAction.role,
+          content: systemPromptAction.content,
+        }
+      }
+    })
+  }
 
-  if (messages.length === 1 && systemPrompt) {
+  if (messages.length <= 2 && systemPrompt) {
```
joelhooks, line 115: we currently pass in the first user message and the title/body of the unsaved content  as an additional system message so if it is 2 or less we know it's an initial message

might make an inngest function that is "chat started" and then have "chat continued" or whatever for future messages to avoid this sort of meh

PR #83, opened 2024-02-27
feat: content metadata to represent content pieces

joelhooks:
one thing i think we might want to avoid is the detachment from "author/instructor" and the user and think about this model from the perspective of a single database where we have a user and a user profile

in sanity this would mean that an "author" is a pointer

another thing i've been considering is the idea of ownership versus authorship and how that affects the model. a resource is "owned by" a user or set of users, but the owners arent necessarily the authors
you've got a set of permissions for any given resource like read/write/managepublished collections (lessons, articles, etc) have sets of contributors (credits), which is different in practice than ownership. a contributor doesn't necessarily have ability to edit/manage the resource for instance.

so there is the ability to manipulate the resources AND the display of credit in the byline that's presented with the resource

i'm not stoked on the term "author" as it's very specific to media type by definition

little gpt sesh to get to this https://chat.openai.com/share/ba3beff6-497b-4bcb-bed6-e93e3412fb0d

One interesting bit is the content_metadata table as a "sanity bridge". Using JSON table makes a ton of sense to me and opens up the possibility of user defined schemas for content types. roles, permissions, content, contribution, etc... all the "types" move to their own tables so they can be defined per-database so we aren't stuck using "tips" etc for every site and they can be shaped accordingly as needed.

content is king

### DIFF
--- a/apps/course-builder-web/src/server/db/schema.ts
+++ b/apps/course-builder-web/src/server/db/schema.ts
@@ -203,8 +203,8 @@ export const rolePermissionsRelations = relations(rolePermissions, ({ one }) => 
 
 export const contentStates = mysqlTable('contentState', {
   id: varchar('id', { length: 255 }).notNull().primaryKey(),
-  slug: varchar('name', { length: 255 }).notNull().unique(),
-  title: varchar('name', { length: 255 }).notNull(),
+  slug: varchar('slug', { length: 255 }).notNull().unique(),
+  name: varchar('name', { length: 255 }).notNull(),
   description: text('description'),
   active: boolean('active').notNull().default(true),
   createdAt: timestamp('createdAt', {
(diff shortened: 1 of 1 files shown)


apps/course-builder-web/src/server/db/schema.ts
```diff
@@ -27,7 +28,7 @@ export const users = mysqlTable(
     id: varchar('id', { length: 255 }).notNull().primaryKey(),
     name: varchar('name', { length: 255 }),
     role: mysqlEnum('role', ['user', 'admin']).default('user'),
-    email: varchar('email', { length: 255 }).notNull(),
+    email: varchar('email', { length: 255 }).notNull().unique(),
```
joelhooks, line 31: this breaks if you ever want to set up a multi-tenant type thing but makes sense for single tenant

PR #86, opened 2024-02-29
chore: setup content migration from sanity to db

joelhooks:
migration


apps/course-builder-web/src/app/page.tsx
```diff
@@ -9,6 +10,9 @@ export const metadata: Metadata = {
 }
 
 export default async function PlaygroundPage() {
+  const resources = await fetchContentResourcesWithTypeAndUserContributions('article')
+
+  console.log(JSON.stringify(resources, null, 2))
```
joelhooks, line 15: demo of loading a resource

apps/course-builder-web/src/app/tips/[slug]/page.tsx
```diff
@@ -23,6 +27,39 @@ type Props = {
...
+    const newTip = TipSchema.parse({
+      _id: newTips[0].id,
+      _updatedAt: newTips[0].updatedAt?.toISOString(),
+      _type: 'explainer',
+      summary: '',
+      ...newTips[0],
+      ...expandedResources[0],
+      ...newTips[0].metadata,
+    })
+  }
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/tips/[slug]/page.tsx:61]: this is also a demo, kind of zany but it'd be transitionary and just showing the same type as a result

apps/course-builder-web/src/inngest/events.ts
```diff
@@ -67,3 +67,10 @@ export type DraftWriteupCompleted = {
     fullPrompt: ChatCompletionRequestMessage[]
   }
 }
+
+export const MIGRATE_EVENT = 'migration'
+
+export type MigrateEvent = {
+  name: typeof MIGRATE_EVENT
+  data: {}
+}
```
joelhooks, line 76: I actually don't know if inngest specifically requires an event to trigger a function manually 🤔

apps/course-builder-web/src/inngest/functions/migration.ts
```diff
@@ -0,0 +1,250 @@
...
+                : null),
+            },
+          })
+
+          console.log('migrating article', sanityArticleResource._id)
+
+          return db.insert(contentResource).values(migratedResource)
+        })
+      }
+    }
```
joelhooks, line 111: this is articles, but videos and tips are below and have the same pattern with a zod schema to validate the shape. in practice we'd adopt these as the schemas for the live resources instead of the flattened structure from sanity which will help us have bespoke metadata and resource structures PER content type which feels nice

apps/course-builder-web/src/inngest/functions/migration.ts
```diff
@@ -0,0 +1,250 @@
...
+    const sanityVideoResources = await step.run('fetch video resources from sanity', async () => {
+      return await sanityQuery<VideoResource[]>(`*[_type == "videoResource"]`, { cache: 'no-cache' })
+    })
+
+    for (const sanityVideoResource of sanityVideoResources) {
+      const canMigrateVideoResource = await step.run('check if video can be migrated', async () => {
+        const existingResource = await fetchContentResource(sanityVideoResource._id)
+
+        return !existingResource
+      })
```
joelhooks, line 122: idempotent ftw

apps/course-builder-web/src/inngest/functions/migration.ts
```diff
@@ -0,0 +1,250 @@
...
+
+          console.log('migrating article', sanityArticleResource._id)
+
+          return db.insert(contentResource).values(migratedResource)
+        })
+      }
+    }
+
+    const sanityVideoResources = await step.run('fetch video resources from sanity', async () => {
+      return await sanityQuery<VideoResource[]>(`*[_type == "videoResource"]`, { cache: 'no-cache' })
```
joelhooks, line 114: added the ability to configure caching because these objects are kind of big with all the transcripts variants, particularly when loading ALL of them

apps/course-builder-web/src/inngest/functions/migration.ts
```diff
@@ -0,0 +1,283 @@
...
+              ...(sanityVideoResource.transcriptWithScreenshots
+                ? [
+                    {
+                      _type: `transcriptWithScreenshots`,
+                      _key: `transcriptWithScreenshots-${nanoid()}`,
+                      text: sanityVideoResource.transcriptWithScreenshots,
+                    },
+                  ]
+                : []),
+            ],
```
joelhooks, line 189: originally had these as metadata, but really transcripts etc are resources and might even be better considered as refs to contentResources because of their volume/weight

here they are not refs

PR #97, opened 2024-03-06
feat: lean into content resource

joelhooks:
lean in


apps/course-builder-web/src/app/articles/_components/edit-article-form.tsx
```diff
@@ -4,10 +4,10 @@ import * as React from 'react'
 import Link from 'next/link'
 import { useRouter } from 'next/navigation'
 import { CodemirrorEditor } from '@/app/_components/codemirror'
-import { updateArticle } from '@/app/articles/_components/article-form-actions'
 import { useIsMobile } from '@/hooks/use-is-mobile'
 import { useSocket } from '@/hooks/use-socket'
 import { Article, ArticleSchema, ArticleVisibilitySchema } from '@/lib/articles'
+import { updateArticle } from '@/lib/articles-query'
```
joelhooks, line 10: these `*-query` files are `'use server'` so need to be "pure" server and can only export functions

apps/course-builder-web/src/app/articles/_components/new-article-form.tsx
```diff
@@ -2,7 +2,8 @@
 
 import * as React from 'react'
 import { useRouter } from 'next/navigation'
-import { createArticle, NewArticle, NewArticleSchema } from '@/app/articles/_components/article-form-actions'
+import { NewArticle, NewArticleSchema } from '@/lib/articles'
```
joelhooks, line 5: these are now types/schemas. it's a little weird, but temporary while thinking about better sorting

apps/course-builder-web/src/app/tips/_components/new-tip-form.tsx
```diff
@@ -34,17 +35,65 @@ export function NewTipForm() {
...
+        yield videoResource
+        return
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, delay))
+      delay += delayIncrement
+    }
+
+    throw new Error('Video resource not found after maximum attempts')
+  }
```
joelhooks, line 58: 🤡 - I wanted to try out a generator

joelhooks: (it does work very well)

we were seeing intermittent issues creating a new tip video because you could hit "create" before the video resources was created in the database after getting uploaded to uploadthing, so now it is polling and doesn't allow a click until it's validated.

the generator using incremental backoff to check if the resource is there and when it is completes and yields the resource.

obviously it sucks because literally .000001% of devs will look at this and think "i see how this works"

apps/course-builder-web/src/db/schema.ts
```diff
@@ -283,10 +284,8 @@ export const contentResource = mysqlTable(
   {
     id: varchar('id', { length: 255 }).notNull().primaryKey(),
     type: varchar('type', { length: 255 }).notNull(),
-    slug: varchar('slug', { length: 255 }).notNull().unique(),
     createdById: varchar('createdById', { length: 255 }).notNull(),
-    resources: json('resources').$type<any[]>().default([]),
```
joelhooks, line 288: `resources` moves to a join table below

apps/course-builder-web/src/db/schema.ts
```diff
@@ -283,10 +284,8 @@ export const contentResource = mysqlTable(
...
-    metadata: json('metadata').$type<Record<string, any>>().default({}),
+    fields: json('fields').$type<Record<string, any>>().default({}),
```
joelhooks, line 288: `metadata` becomes `fields`

moves `slug` in here since it isn't universal.

apps/course-builder-web/src/db/schema.ts
```diff
@@ -310,6 +309,47 @@ export const contentResource = mysqlTable(
...
+    fields: [contentResourceResource.resourceOfId],
+    references: [contentResource.id],
+    relationName: 'resourceOf',
+  }),
+  resource: one(contentResource, {
+    fields: [contentResourceResource.resourceId],
+    references: [contentResource.id],
+    relationName: 'resource',
+  }),
 }))
```
joelhooks, line 353: this gives us a join table for parent/child relationships with `contentResources` effectively making them collections. Resources get a `position` so we can sort.

PR #103, opened 2024-03-13
feat: add concept store

mbilokonsky:
This is still a work in progress but wanted to show what I've got.


apps/course-builder-web/src/inngest/functions/concepts/get-or-create-tag.ts
```diff
@@ -0,0 +1,30 @@
...
+    })
+
+    // TODO:
+    // 1. check for similarity, we need to tune this.
+    // 2. if similarity is found, broadcast a REQUEST_CONCEPT_SELECTION event
+    // 3. if no similarity is found, create a new concept and broadcast a REQUEST_CONCEPT_SELECTION event
+
+    return { new_concept, related_concepts }
+  },
+)
```
joelhooks, line 30:
```suggestion
  async ({ event, step }) => {
    const related_concepts = await step.run('check for similar concepts', async () => {
      return await get_related_concepts(event.data.key)
    })

    const new_concept = await step.run('create new concept', async () => {
      return await add_concept(event.data.key)
    })

    // TODO:
    // 1. check for similarity, we need to tune this.
    // 2. if similarity is found, broadcast a REQUEST_CONCEPT_SELECTION event
    // 3. if no similarity is found, create a new concept and broadcast a REQUEST_CONCEPT_SELECTION event

    return { new_concept, related_concepts }
  },
)
```

With inngest you typically want to store return values not set values outside of the immediate scope of the function

apps/course-builder-web/src/inngest/functions/split_video.ts
```diff
@@ -0,0 +1,23 @@
...
+  {
+    id: `compute split points`,
+    name: 'Compute Split Points',
+  },
+  {
+    event: REQUEST_VIDEO_SPLIT_POINTS,
+  },
+  async ({ event, step }) => {
+    const split_points = await step.run('determining split points', async () => {
+      determineSplitPoints(event.data.transcript)
```
joelhooks, line 16:
```suggestion
      return determineSplitPoints(event.data.transcript)
```

needs to return

apps/course-builder-web/src/inngest/functions/split_video.ts
```diff
@@ -0,0 +1,23 @@
...
+    })
+
+    // open question - do we want to return the split poinst here, or do we want to kick off an actual splitting of the video within this function?
+    // feels like maybe we want to dispatch another event, onoe that maps split points to event.data.resource, and have that mapped to some sort of actual video splitting behavior?
```
joelhooks, line 20: "persist and notify" would be sensible steps here

splits would be a field on the video resource and dispatching a "splits are ready" event makes a ton of sense to me

joelhooks (approved):

PR #121, opened 2024-03-23
chore: add relations to the drizzle adapter

joelhooks:
relations


apps/course-builder-web/src/db/schema.ts
```diff
@@ -93,4 +32,17 @@ export {
...
+	contentContributions,
+	contentContributionRelations,
+	contentResource,
+	contentResourceRelations,
+	contentResourceResource,
+	contentResourceResourceRelations,
+	contributionTypes,
+	contributionTypesRelations,
+	resourceProgress,
+} = getCourseBuilderSchema(mysqlTable)
```
joelhooks, line 48: nice clean little call

the cool bit here is that consumer apps can both pick and choose which bits to add, but can easily extend or add new tables to fill any needs they may have that aren't met with these defaults 🌀

PR #126, opened 2024-03-25
feat: adds concept of sessions and callbacks to Course Builder

joelhooks:
callback


apps/course-builder-web/src/app/page.tsx
```diff
@@ -9,6 +10,8 @@ export const metadata: Metadata = {
 }
 
 export default async function PlaygroundPage() {
+	const cb = await coursebuilder()
+	console.log({ cb })
```
joelhooks, line 14: just a demo

apps/course-builder-web/src/coursebuilder/course-builder-config.ts
```diff
@@ -1,26 +1,22 @@
...
+	adapter: courseBuilderAdapter,
 	inngest,
 	providers: [transcriptProvider],
 	basePath: '/api/coursebuilder',
+	callbacks: {
+		session: async (req) => {
+			// TODO: there's nothing on the "session" but we can add whatever we want here
+			return { ...req, example: 'just an example' }
+		},
+	},
```
joelhooks, line 19: "session" is probably the wrong term, but it also makes OK sense. we can add anything we need here eventually

* progress
* personalization
* feature flags/entitlements
* ???

packages/core/src/lib/actions/index.ts
```diff
@@ -1,6 +1,27 @@
...
+	const { callbacks, logger } = options
+
+	const response: ResponseInternal<any | null> = {
+		body: null,
+		headers: { 'Content-Type': 'application/json' },
+		cookies,
+	}
+
+	try {
+		response.body = await callbacks.session({})
```
joelhooks, line 17: we have access to the database adapter here, so anything available on that should be here

we aren't doing anything to tie into the current user session, so we want to consider that

one idea I had was a `getAuth` or similar on options that would return an AdapterSession/AsapterUser

still noodling

packages/next/tsconfig.json
```diff
@@ -10,8 +10,8 @@
 		"experimentalDecorators": true,
 		"forceConsistentCasingInFileNames": true,
 		"isolatedModules": true,
-		"module": "ESNext",
-		"moduleResolution": "node",
+		"module": "NodeNext",
+		"moduleResolution": "NodeNext",
```
joelhooks, line 14: https://x.com/jhooks/status/1772366626655990090?s=20

PR #139, opened 2024-04-05
feat: migrate the Skill Stack prisma api

joelhooks:
this is the start of a journey

https://www.loom.com/share/f2f72833b43c44f5bdb6d8b6a7c10861

journey


apps/pro-aws/src/app/admin/page.tsx
```diff
@@ -1,12 +1,93 @@
...
+			return
+		}
+
+		await db.insert(userRoles).values({
+			roleId: adminRole.id,
+			userId: user.id,
+		})
+
+		revalidatePath('/admin')
+	}
```
joelhooks, line 73: i gotta say, server components are wild

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -163,9 +188,316 @@ export function mySqlDrizzleAdapter(
...
+			purchases: any,
+			productId: string,
+		): Promise<
+			{
+				upgradableTo: { id: string; name: string }
+				upgradableFrom: { id: string; name: string }
+			}[]
+		> {
+			return Promise.resolve([])
+		},
```
joelhooks, line 205: most of these are still noop and need to be implemented

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -163,9 +188,316 @@ export function mySqlDrizzleAdapter(
...
+			)
+
+			return Promise.resolve({
+				availableUpgrades: z
+					.array(upgradableProductSchema)
+					.parse(availableUpgrades),
+				existingPurchase,
+				purchase: parsedPurchase.data,
+			})
+		},
```
joelhooks, line 414: implementing the more challenging methods first

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -163,9 +188,316 @@ export function mySqlDrizzleAdapter(
...
+				),
+				with: {
+					user: true,
+					product: true,
+					bulkCoupon: true,
+				},
+				orderBy: asc(purchase.createdAt),
+			})
+
+			const parsedPurchases = z.array(purchaseSchema).safeParse(userPurchases)
```
joelhooks, line 458: using zod

one thing we don't have is type safety flowing through in the adapter when using the drizzle client

something to look in to, but even then using zod to parse them into known types works well and also gives us schemas/types for the core stuff

packages/core/src/adapters.ts
```diff
@@ -43,3 +256,130 @@ export const MockCourseBuilderAdapter: CourseBuilderAdapter = {
...
+	>
+	updatePurchaseUserTransferTransferState(options: {
+		id: string
+		transferState: PurchaseUserTransferState
+	}): Promise<PurchaseUserTransfer | null>
+	pricesOfPurchasesTowardOneBundle(options: {
+		userId: string | undefined
+		bundleId: string
+	}): Promise<Price[]>
+}
```
joelhooks, line 385: the full interface for the Skill Stack "prisma api" which is used a lot, but not completely

packages/utils/adapter.ts
```diff
@@ -153,6 +154,20 @@ export async function runBasicTests(options: TestOptions) {
...
+			productId: 'fake-product-id',
+			merchantChargeId: 'fake-charge-id',
+			merchantSessionId: 'fake-session-id',
+			totalAmount: '123',
+		})
+
+		const parsedPurchase = purchaseSchema.safeParse(purchase)
+
+		expect(parsedPurchase.success).toBeTruthy()
+	})
```
joelhooks, line 169: tests against a real db should be nice

these tests will work against any of the db flavors and aren't specific to the mysql implementation

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -161,11 +193,503 @@ export function mySqlDrizzleAdapter(
...
+					throw error
+				}
+			})
+
+			return purchaseSchema.parse(
+				await client.query.purchases.findFirst({
+					where: eq(purchaseTable.id, purchaseId as string),
+				}),
+			)
+		},
```
joelhooks, line 430: this is one of the most gnarly ones, but I think the transaction should just work here

noodling on how to test

packages/adapter-drizzle/test/mysql/schema.ts
```diff
@@ -37,9 +37,10 @@ export const {
 	merchantPrice,
 	merchantProduct,
 	merchantSession,
-	price,
-	product,
-	purchase,
+	prices,
+	products,
+	purchases,
```
joelhooks, line 42: annoying that this isn't all consistent case 😭

PR #150, opened 2024-04-07
feat: setup a package for rsc client components

joelhooks:
boxes


apps/course-builder-web/src/app/(content)/tips/_components/create-tip.tsx
```diff
@@ -1,13 +1,33 @@
...
+							`/${pluralize(resource.type)}/${resource.fields?.slug || resource.id}/edit`,
+						)
+					}}
+					createResource={createTip}
+					getVideoResource={getVideoResource}
+				>
+					{(handleSetVideoResourceId: (id: string) => void) => {
+						return <TipUploader setVideoResourceId={handleSetVideoResourceId} />
+					}}
+				</NewResourceWithVideoForm>
```
joelhooks, line 30: this is more verbose here, but it moves a huge component to a shared package and leaves us with config details to inject versus the entire component

apps/course-builder-web/src/app/(content)/tips/_components/create-tip.tsx
```diff
@@ -1,13 +1,33 @@
...
+						router.push(
+							`/${pluralize(resource.type)}/${resource.fields?.slug || resource.id}/edit`,
+						)
+					}}
+					createResource={createTip}
+					getVideoResource={getVideoResource}
+				>
+					{(handleSetVideoResourceId: (id: string) => void) => {
+						return <TipUploader setVideoResourceId={handleSetVideoResourceId} />
+					}}
```
joelhooks, line 29: the old "function as child" approach let's us compose using the internal call backs

`TipLoader` is using UploadThing which relies on "application context" so it's more difficult to yoink out

can we just use s3 at this point? our needs are simple.

packages/react-rsc/package.json
```diff
@@ -0,0 +1,57 @@
...
+		".": {
+			"types": "./dist/index.d.ts",
+			"import": "./dist/index.js",
+			"require": "./dist/index.cjs"
+		},
+		"./client": {
+			"types": "./dist/client.d.ts",
+			"import": "./dist/client.js",
+			"require": "./dist/client.cjs"
+		}
```
joelhooks, line 32: this is a client specific entry point

dunno if this will properly package server actions yet

PR #152, opened 2024-04-08
chore: migrate and test format prices for product

joelhooks:
brings the `formatPricesForProduct` method into Coures Builder from Skill Stack

with Skill Stack we assume prisma and have a direct ORM level connection to the database

for tests we used a deep mock via jest, but in this case we don't have that and want to

* support multiple database adapters
* run tests against a real database

the adapter constraint means that we can't rely on ORM or database direct syntax and need to leverage our existing adapters. this means that our tests are ran in the context of the adapter, and not the core library

the testing harness was already set up in the `utils` package for running authjs adapter tests, so here i added the stack of price formatting tests, commented out the most broken ones, and got the first 4 simple tests running

now we setup and teardown products in the database, and have a clear/simple way to address the data access via the `CourseBuilderAdapter` as well as a little test harness adapter that we can use to add test only methods with direct orm/db/sql access

👋🏻 note: this PR removes the sqlite and postgres adapters for now

inspection


apps/course-builder-web/src/app/(content)/events/[slug]/page.tsx
```diff
@@ -0,0 +1,200 @@
...
+	totalQuantity: number
+	purchaseCount: number
+	product?: Product
+	mdx?: MDXRemoteSerializeResult
+	hasPurchasedCurrentProduct?: boolean
+	availableBonuses: any[]
+	existingPurchase?: Purchase & { product?: Product | null }
+	purchases?: Purchase[]
+	userId?: string
+} & CommerceProps
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/(content)/events/[slug]/page.tsx:48]: goal is to get an events page up that will allow ticket sales and have the end to end commerce flows in place

apps/course-builder-web/src/app/(content)/events/[slug]/page.tsx
```diff
@@ -0,0 +1,200 @@
...
+		}
+	} else {
+		eventProps = {
+			event,
+			availableBonuses: [],
+			quantityAvailable: -1,
+			totalQuantity: -1,
+			purchaseCount: 0,
+		}
+	}
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/(content)/events/[slug]/page.tsx:172]: "brute force" approach here, loading all the data that is in `getServerSideProps`

apps/course-builder-web/src/lib/props-for-commerce.ts
```diff
@@ -0,0 +1,114 @@
...
+								? Number(purchase.totalAmount)
+								: purchase.totalAmount,
+					}
+				}),
+			],
+		}),
+		products,
+		allowPurchase,
+	}
+}
```
joelhooks, line 114: classic `propsForCommerce` still puttin in that work

apps/course-builder-web/src/app/(content)/events/[slug]/_components/event-template.tsx
```diff
@@ -0,0 +1,920 @@
...
+	// 			couponId,
+	// 			merchantCoupon,
+	// 			autoApplyPPP,
+	// 		},
+	// 		{
+	// 			onSuccess: (formattedPrice) => {
+	// 				addPrice(formattedPrice, productId)
+	// 			},
+	// 		},
+	// 	)
```
Comment (joelhooks) [inline apps/course-builder-web/src/app/(content)/events/[slug]/_components/event-template.tsx:327]: eventually you get here to `pricing-formatted` that calls `formatPricesForProduct` and gives use:

```
export type FormattedPrice = {
	id: string
	quantity: number
	unitPrice: number
	fullPrice: number
	fixedDiscountForUpgrade: number
	calculatedPrice: number
	availableCoupons: Array<
		Omit<MerchantCouponWithCountry, 'identifier'> | undefined
	>
	appliedMerchantCoupon?: MinimalMerchantCoupon
	upgradeFromPurchaseId?: string
	upgradeFromPurchase?: Purchase
	upgradedProduct?: ProductWithPrices | null
	bulk: boolean
	usedCouponId?: string
}
```

that's a huge chunk of logic for commerce

well tested, but...

packages/utils/format-prices-for-product.test.ts
```diff
@@ -0,0 +1,891 @@
...
+	// 			getMockCoupon('bulk', getBulkDiscountPercent(quantity)),
+	// 		])
+	//
+	// 		mockCtx.merchantCoupon.findFirst.mockResolvedValue(MOCK_SITE_SALE_COUPON)
+	//
+	// 		const { expectedPrice, calculatedPrice } =
+	// 			await expectedPriceForDefaultCoupon(quantity)
+	// 		expect(expectedPrice).toBe(calculatedPrice)
+	// 	})
+	// }
```
joelhooks, line 79: these tests all assume you've got a fully mocked and typed prisma client

we don't

we have an adapter that doesn't leak database or orm details to the consumers

that's great, but makes things more interesting

we need to be able to **run these tests** against any adapter, and let the adapter handle the implementation details

packages/adapter-drizzle/test/mysql/index.test.ts
```diff
@@ -85,3 +90,104 @@ runBasicTests({
...
+				.delete(prices)
+				.where(eq(prices.productId, productId))
+				.then((res) => res[0])
+			await db
+				.delete(merchantCoupon)
+				.where(eq(merchantCoupon.id, fixtures?.coupon.id))
+				.then((res) => res[0])
+		},
+	},
+})
```
joelhooks, line 193: this calls the `runFormatPricingTests` that lives in the utils package and injects context, including the adapter and a little mini `db` adapter for the test runner

packages/adapter-drizzle/test/mysql/index.test.ts
```diff
@@ -85,3 +90,104 @@ runBasicTests({
...
+				.then((res) => res[0])
+			await db
+				.delete(prices)
+				.where(eq(prices.productId, productId))
+				.then((res) => res[0])
+			await db
+				.delete(merchantCoupon)
+				.where(eq(merchantCoupon.id, fixtures?.coupon.id))
+				.then((res) => res[0])
+		},
```
joelhooks, line 191: here we can create/destroy the "product stack" 

what's rad here is that we are working against a live mysql database so nothing is mocked, it's all integrated with the actual database

this means the tests take a bit longer to run because it starts the database everytime

packages/utils/adapter.ts
```diff
@@ -14,6 +14,13 @@ export interface TestOptions {
 		sessionUpdateExpires?: Date
 		verificationTokenExpires?: Date
 		createdAt?: Date
+		product?: any
+		upgradeProduct?: any
+		price?: any
+		upgradePrice?: any
+		coupon?: any
+		largeCoupon?: any
+		indiaCoupon?: any
```
joelhooks, line 23: new fixtures for commerce tests

packages/utils/adapter.ts
```diff
@@ -47,6 +54,15 @@ export interface TestOptions {
...
 		authenticator?: (credentialID: string) => any
+		createProduct?: (product: {
+			id: string
+			name: string
+			createdAt: Date
+			key: string
+			status: number
+			quantityAvailable: number
+		}) => any
+		deleteProduct?: (productId: string) => any
```
joelhooks, line 65: product entities for the test adapter

packages/utils/format-prices-for-product.test.ts
```diff
@@ -0,0 +1,891 @@
...
+export async function runFormatPricingTests(options: TestOptions) {
+	let ctx: CourseBuilderAdapter = options.adapter
+
+	beforeEach(async () => {
+		await options.db.createProduct?.(options.fixtures?.product)
+	})
+
+	afterEach(async () => {
+		await options.db.deleteProduct?.(options.fixtures?.product?.id)
+	})
```
joelhooks, line 30: setting up and tearing down products before each run

packages/utils/format-prices-for-product.test.ts
```diff
@@ -0,0 +1,891 @@
...
+		})
+		const expectedPrice = product?.unitPrice * quantity
+		expect(expectedPrice).toBe(product?.calculatedPrice)
+	})
+
+	test('applies sale coupon as default', async () => {
+		const { expectedPrice, calculatedPrice } =
+			await expectedPriceForDefaultCoupon()
+		expect(expectedPrice).toBe(calculatedPrice)
+	})
```
joelhooks, line 65: these 4 pass ✅🎉

PR #155, opened 2024-04-16
feat: pricing widget 

joelhooks:
checkout


apps/course-builder-web/src/coursebuilder/course-builder-config.ts
```diff
@@ -14,6 +15,7 @@ const stripeProvider = StripeProvider({
 	errorRedirectUrl: `${env.COURSEBUILDER_URL}/checkout-error`,
 	baseSuccessUrl: `${env.COURSEBUILDER_URL}/checkout-success`,
 	cancelUrl: `${env.COURSEBUILDER_URL}/checkout-cancel`,
+	paymentsAdapter: new StripePaymentAdapter(),
```
joelhooks, line 18: this is a touch clunky since we have an adapter that takes a provider that takes an adapter, but in practice it does the job very well and allows us to limit the surface area of the stripe client into a single interface and concrete implementation

apps/course-builder-web/src/coursebuilder/stripe-payment-adapter.ts
```diff
@@ -0,0 +1,44 @@
...
+	async createCustomer(params: { email: string; userId: string }) {
+		const stripeCustomer = await this.stripe.customers.create({
+			email: params.email,
+			metadata: {
+				userId: params.userId,
+			},
+		})
+		return stripeCustomer.id
+	}
+}
```
joelhooks, line 44: a nice wrapper around stripe with "domain specific needs"

it fits the "adapter" pattern we've been using higher up and (for me anyway) it's easy to understand

packages/core/src/lib/pricing/stripe-checkout.ts
```diff
@@ -71,18 +73,18 @@ async function findOrCreateStripeCustomerId(
 				provider: 'stripe',
 			})
 			if (merchantAccount) {
-				const customer = await stripe.customers.create({
+				const customerId = await paymentsAdapter.createCustomer({
```
joelhooks, line 76: this is the good bit! now "stripe checkout" doesn't care about actually communicating with stripe, it just asks the adapter for what it needs and doesn't even get the whole object back, just the identifier. nice.

packages/core/src/lib/pricing/stripe-checkout.ts
```diff
@@ -308,16 +306,19 @@ export async function stripeCheckout({
...
+					? await config.paymentsAdapter.getCouponPercentOff(
+							merchantCoupon.identifier,
+						)
 					: 0
 
+			// const stripeCouponPercentOff =
+			// 	stripeCoupon && stripeCoupon.percent_off
+			// 		? stripeCoupon.percent_off / 100
+			// 		: 0
+			// TODO: be sure to divide it by 100 in the method🤡
```
joelhooks, line 320: I did this ;)

packages/core/src/providers/stripe.ts
```diff
@@ -12,11 +14,42 @@ export interface StripeProviderConfig {
...
+	/**
+	 * Returns the URL to redirect to for a checkout session.
+	 * @param params
+	 */
+	createCheckoutSession(
+		params: Stripe.Checkout.SessionCreateParams,
+	): Promise<string>
+	createCustomer(params: Stripe.CustomerCreateParams): Promise<string>
+}
+
```
joelhooks, line 46: we can add to this as needed for other stripe stuff we might want product sync, etc)

apps/course-builder-web/src/coursebuilder/stripe-payment-adapter.ts
Creeland: I like how this is organized

apps/course-builder-web/src/coursebuilder/course-builder-config.ts
joelhooks: the clunkiness is jsut how things are named, but nesting is probably fine?

packages/core/src/lib/pricing/stripe-checkout.ts
Creeland: That is nice. Only exposing what's needed

packages/core/src/lib/pricing/stripe-checkout.ts
```diff
@@ -308,16 +306,19 @@ export async function stripeCheckout({
...
+				merchantCoupon && merchantCoupon.identifier
+					? await config.paymentsAdapter.getCouponPercentOff(
+							merchantCoupon.identifier,
+						)
 					: 0
 
+			// const stripeCouponPercentOff =
+			// 	stripeCoupon && stripeCoupon.percent_off
+			// 		? stripeCoupon.percent_off / 100
+			// 		: 0
```
Creeland, line 319: So now checkout doesn't need to be concerned with any coupon logic anymore 👍

PR #157, opened 2024-04-23
feat(proaws): stub out video player overlays rendering logic

vojtaholik:
screenshot

gif


packages/ui/hooks/use-video-player-overlay.tsx
joelhooks, line 1: i'm not super stoked about continuing to expand the surface area of the UI package unless absolutely required

it's a PITA

joelhooks (approved):

PR #165, opened 2024-04-26
fix(proaws): tutorial related user abilities

vojtaholik:
```tsx
// example
const { data: abilityRules } = api.ability.getCurrentAbilityRules.useQuery({
  lessonId: params.lesson,
  moduleId: params.module,
})
const ability = createAppAbility(abilityRules || [])
const canView = ability.can('read', 'Content')
```

gif

### DIFF
--- a/apps/pro-aws/src/ability.ts
+++ b/apps/pro-aws/src/ability.ts
@@ -7,7 +7,7 @@ import {
 } from '@casl/ability'
 import z from 'zod'
 
-import { ContentResourceSchema } from '@coursebuilder/core/schemas/content-resource-schema'
+import { ContentResourceResourceSchema } from '@coursebuilder/core/schemas/content-resource-schema'
 import { ContentResource } from '@coursebuilder/core/types'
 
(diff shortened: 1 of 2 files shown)


apps/pro-aws/src/ability.ts
```diff
@@ -122,11 +122,6 @@ export function defineRulesForPurchases(
 		can(['read', 'update'], 'User', { id: user.id })
 	}
 
-	can('read', 'Content', {
-		createdAt: { $lte: new Date() },
-		status: { $in: ['review', 'published'] },
-	})
-
```
vojtaholik, line 129: I'm not exactly sure what this was for, but I don't think we need it anymore since it was basically allowing anyone to "read".

joelhooks: the created at is anything in the past (ie not scheduled) or if the status is "review" (not sure) or "published" (generally available) but not any other state

probably still valid

joelhooks (approved):

PR #167, opened 2024-04-26
feat: add the product pricing widget

joelhooks:
wedge

### DIFF
--- a/apps/course-builder-web/src/app/(commerce)/products/[slug]/_components/product-pricing.tsx
+++ b/apps/course-builder-web/src/app/(commerce)/products/[slug]/_components/product-pricing.tsx
@@ -1,0 +1,39 @@
+'use client'
+
+import * as React from 'react'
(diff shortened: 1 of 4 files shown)


packages/core/src/inngest/commerce/send-post-purchase-email.ts
```diff
@@ -0,0 +1,62 @@
...
+			email: user.email as string,
+			callbackUrl: `${siteRootUrl}/welcome?purchaseId=${purchase.id}`,
+			baseUrl: siteRootUrl,
+			authOptions: getAuthConfig(),
+			emailProvider: emailProvider,
+			adapter: db,
+			merchantChargeId: purchase.merchantChargeId,
+		})
+	})
+}
```
joelhooks, line 56: breaking this off into its own function makes testing way easier since it can be re-ran independently

PR #168, opened 2024-05-01
feat: continue migrating the pricing widget

joelhooks:
make money


apps/course-builder-web/src/pricing/pricing.tsx
```diff
@@ -70,25 +74,35 @@ export const Pricing: React.FC<React.PropsWithChildren<PricingProps>> = ({
...
+			productId,
+			quantity: debouncedQuantity,
+			couponId,
+			merchantCoupon,
+			autoApplyPPP,
+		},
+		{
+			...(initialFormattedPrice && { initialData: initialFormattedPrice }),
+		},
+	)
```
joelhooks, line 105: the `pricingDataLoader` gives us the initial price and then we pick it back up with trpc for dynamic pricing if they make changes (ppp, team, etc)

one thing we haven't done is affect the query params, which would make a nice addition for team so the quantity would persist in links/reloads

PR #178, opened 2024-05-13
fix(proaws): safely parse user in module progress getter

vojtaholik:
I was getting parsing errors due to `roles` field on user object not returning expected shape defined by `userSchema`.

```tsx
const user = await client.query.users.findFirst({
	where: or(eq(users.id, userIdOrEmail), eq(users.email, userIdOrEmail)),
	with: {
		roles: true,
	},
})

// returns:
user {
   ...,
   name: 'Vojta Holik',
   role: 'user',
   roles: [
     {
       userId: 'ABC',
       roleId: 'role_ABC',
       active: true,
       createdAt: 2024-04-08T23:39:34.000Z,
       updatedAt: 2024-04-08T23:39:34.000Z,
       deletedAt: null
     },
     {
       userId: 'ABC',
       roleId: 'role_XYZ',
       active: true,
       createdAt: 2024-04-08T23:45:38.000Z,
       updatedAt: 2024-04-08T23:45:38.000Z,
       deletedAt: null
     }
   ]
 }
```

while `user-schema` has following expectations:

```
...,
roles: z
	.array(
		z.object({
			id: z.string(),
			name: z.string(),
			description: z.string().nullable(),
			active: z.boolean(),
		}),
	)
	.optional()
	.default([]),
```

`ability` file is expecting above shape so instead of fixing the types I chose to omit the roles field since it wasn't used in the getter function.

gif


packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -986,9 +986,6 @@ export function mySqlDrizzleAdapter(
 
 			const user = await client.query.users.findFirst({
 				where: or(eq(users.id, userIdOrEmail), eq(users.email, userIdOrEmail)),
-				with: {
-					roles: true,
-				},
```
vojtaholik, line 991: I haven't tried fixing the types yet; instead, I chose to omit the roles since they weren't used in this getter function.

joelhooks: the problem is that `roles` is a join table that has a reference `role` so if you use `adapter.getUserById` it returns the right shape. Challenge here is we also use email so you could look at what that's doing and build the user in a similar way.

vojtaholik: I see, I modified the user query per `getUserById` 👍

PR #179, opened 2024-05-13
feat: display Intl formatted time zone identifier

jbranchaud:
@joelhooks I was just reading through your most recent PR and was wondering if we could format the displayed time zone identifier with the Intl API. Turns out to be pretty easy and has full browser support.

Not sure if we want to always display events in terms of Pacific Time or if that changes on an event-by-event basis.

I was also curious about the `timezone` value being destructured from fields. What does that value look like and should we use that for a dynamic time zone of the event?

This screenshot demos the behavior:

CleanShot 2024-05-13 at 09 41 01@2x

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

sundial


joelhooks:
I'm typically trying to stick to Pacific for all times https://github.com/badass-courses/course-builder/blob/54fbdd1b23ab40e923bb1027c2634b8ebe8a30cc/apps/course-builder-web/src/app/(content)/events/%5Bslug%5D/edit/_components/edit-event-form.tsx#L138-L152

PR #181, opened 2024-05-16
feat: add a slack provider

joelhooks:
slack


apps/pro-aws/src/app/(commerce)/thanks/purchase/page.tsx
```diff
@@ -228,6 +225,7 @@ export default async function ThanksPurchasePage({
 						cancelPurchaseTransfer={cancelPurchaseTransfer}
 						initiatePurchaseTransfer={initiatePurchaseTransfer}
 					>
+						<PurchaseTransfer.Header />
```
joelhooks, line 228: noticed this showed even with no transfer so updated with header

PR #185, opened 2024-05-25
chore: React 19 RC Upgrade

joelhooks:
upgrading


joelhooks:
gonna table this for another time when there are stable versions out

PR #189, opened 2024-06-11
feat: (value-based) add initial landing page copy

zacjones93:
screenshot

gif


apps/value-based-design/src/app/page.tsx
```diff
@@ -18,109 +18,166 @@ export default async function Home() {
 				<h1 className="font-heading text-center text-5xl font-semibold sm:text-6xl lg:text-7xl xl:text-8xl">
 					Value-Based Design
 				</h1>
```
zacjones93, line 20: the H1 could be iterated on, telling us who this product is for. 

Had some ideas but none of them feel right...

Make valuable impact on your business through design
Move the needle for your business through your design
Become indispensible because of your design decisions
Link increased business revenue to your design decisions
Become essntial through your design at your company

vojtaholik (approved):

PR #191, opened 2024-06-13
feat(pronextjs): og image previews

vojtaholik:
screenshot

gif


apps/pro-nextjs/src/app/(content)/workshops/page.tsx
```diff
@@ -20,6 +21,13 @@ import {
 
 export const metadata: Metadata = {
 	title: `Pro Next.js Workshops by ${config.author}`,
+	openGraph: {
+		images: [
+			{
+				url: `${env.NEXT_PUBLIC_URL}/api/og?title=${encodeURIComponent(`Pro Next.js Workshops by ${config.author}`)}`,
```
joelhooks, line 27: this should be a method. the same one, probably with a type like `ContentResource | string` where if it's a string it gives the simple text card

joelhooks (approved):

PR #193, opened 2024-06-13
feat: use a generic Module type instead of Workshop or Tutorial 

joelhooks:
modules have a specific shape, so we can type them as such and they should be the same between tutorials and workshops.

loom tour: https://www.loom.com/share/0f7333c1aecd49adb7e7497f78eb20be?sid=6e6825c1-3325-4c49-a0cb-8b138d2b4046

shapes


apps/pro-nextjs/src/ability.ts
```diff
@@ -223,7 +224,7 @@ const isFreelyVisible = ({
 	}
 
 	const lessons = z
-		.array(ContentResourceResourceSchema)
+		.array(z.object({ resourceId: z.string() }))
```
joelhooks, line 227: below we only care that this one property exists so let's check for that instead of an entire tree that is very specific

apps/pro-nextjs/src/app/(content)/_components/video-player-overlay.tsx
```diff
@@ -22,7 +23,7 @@ import type { CompletedAction } from '@coursebuilder/ui/hooks/use-video-player-o
 export const CompletedLessonOverlay: React.FC<{
 	action: CompletedAction
 	resource: ContentResource | null
-	moduleResource: ContentResource | null
+	moduleResource: Module | null
```
joelhooks, line 26: a Module is more specific shape of content resource that has fields and resources we expect

apps/pro-nextjs/src/app/(content)/tips/[slug]/page.tsx
```diff
@@ -87,28 +87,12 @@ export default async function TipPage({
...
-						<Suspense
-							fallback={
-								<div className="px-5 sm:px-8">
-									<Skeleton className="h-16 w-full bg-gray-100" />
-								</div>
-							}
-						>
-							<TipBody tipLoader={tipLoader} />
-						</Suspense>
+						<TipBody tipLoader={tipLoader} />
```
Comment (joelhooks) [inline apps/pro-nextjs/src/app/(content)/tips/[slug]/page.tsx:90]: i don't think we need/want suspense boundaries around all of these things

apps/pro-nextjs/src/app/(content)/tutorials/[module]/[lesson]/(view)/page.tsx
```diff
@@ -67,24 +69,47 @@ export async function generateMetadata(
...
+				<LessonPage
+					lessonLoader={lessonLoader}
+					exerciseLoader={exerciseLoader}
+					moduleLoader={tutorialModuleLoader}
+					moduleProgressLoader={moduleProgressLoader}
+				/>
+			</LessonProvider>
+		</ModuleProvider>
+	)
+}
```
Comment (joelhooks) [inline apps/pro-nextjs/src/app/(content)/tutorials/[module]/[lesson]/(view)/page.tsx:100]: all the loaders, also playing with a couple providers for the deeper components like completion toggle (which I haven't updated yet)

apps/pro-nextjs/src/app/(content)/tutorials/[module]/[lesson]/(view)/page.tsx
```diff
@@ -67,24 +69,47 @@ export async function generateMetadata(
 
 export type Props = {
 	params: { lesson: string; module: string }
-	isSolution?: boolean
-	isExercise?: boolean
+	lessonPageType?: 'exercise' | 'solution' | 'default'
```
Comment (joelhooks) [inline apps/pro-nextjs/src/app/(content)/tutorials/[module]/[lesson]/(view)/page.tsx:72]: using a page type instead of booleans ftw

vojtaholik (approved):
⭐

PR #202, opened 2024-07-10
feat(vbd) add workshops routes for value-based design

zacjones93:
This adds workshops to vbd largely following structure of pronextjs

gif


zacjones93:
Not sure how to address the linting error: 

```
☔️ error course-builder-turbo has a dependency on @coursebuilder/eslint-config@0.0.4 but the version of @coursebuilder/eslint-config in the repo is 0.1.0 which is not within range of the depended on version, please update the dependency version
```

Creeland (approved):
Looks good. I don't see any obvious issues

PR #204, opened 2024-07-19
fix(pnjs): workshop saving

vojtaholik:
gif

### DIFF
--- a/apps/pro-nextjs/src/components/lesson-list/tree.tsx
+++ b/apps/pro-nextjs/src/components/lesson-list/tree.tsx
@@ -97,7 +97,7 @@ export default function Tree({
 			if (item.children.length > 0) {
 				for (const childItem of item.children) {
 					if (!childItem.itemData) continue
-					await updateResourcePosition({
+					return await updateResourcePosition({
 						currentParentResourceId: childItem.itemData.resourceOfId,
 						parentResourceId: item.itemData.resourceId,
 						resourceId: childItem.itemData.resourceId,
@@ -105,7 +105,7 @@ export default function Tree({
 					})
 				}
 			}
-			await updateResourcePosition({
+			return await updateResourcePosition({
 				currentParentResourceId: item.itemData.resourceOfId,
 				parentResourceId: rootResourceId,
 				resourceId: item.itemData.resourceId,
@@ -116,6 +116,10 @@ export default function Tree({
 
 	useEffect(() => {
 		if (lastAction === null) {
+			return
+		}
+
+		if (lastAction.type === 'toggle') {
 			return
 		}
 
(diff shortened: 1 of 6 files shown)


apps/pro-nextjs/src/components/lesson-list/tree.tsx
```diff
@@ -97,15 +97,15 @@ export default function Tree({
...
+					return await updateResourcePosition({
 						currentParentResourceId: childItem.itemData.resourceOfId,
 						parentResourceId: item.itemData.resourceId,
 						resourceId: childItem.itemData.resourceId,
 						position: item.children.indexOf(childItem),
 					})
 				}
 			}
-			await updateResourcePosition({
+			return await updateResourcePosition({
```
vojtaholik, line 108: return here to avoid looping forever

apps/pro-nextjs/src/components/lesson-list/tree.tsx
```diff
@@ -119,6 +119,10 @@ export default function Tree({
 			return
 		}
 
+		if (lastAction.type === 'toggle') {
```
vojtaholik, line 122: no saving necessary if we are just toggling section open

packages/ui/resources-crud/edit-resources-form-desktop.tsx
```diff
@@ -73,10 +73,8 @@ export function EditResourcesFormDesktop({
 	theme?: string
 }) {
 	const onSubmit = async (values: z.infer<typeof resourceSchema>) => {
-		const updatedResource = await updateResource(values)
-		if (updatedResource) {
-			onSave(updatedResource)
-		}
```
vojtaholik, line 79: `updatedResource` comes back with all its resources which can exceed server action size limit.

packages/ui/resources-crud/edit-resources-form-desktop.tsx
```diff
@@ -75,7 +75,7 @@ export function EditResourcesFormDesktop({
 	const onSubmit = async (values: z.infer<typeof resourceSchema>) => {
 		const updatedResource = await updateResource(values)
 		if (updatedResource) {
-			onSave(updatedResource)
+			return await onSave(updatedResource)
 		}
 	}
```
vojtaholik, line 80: image


I tracked the error down to this onSubmit function where we return `updatedResource` with all its `resources` into server function

apps/pro-nextjs/src/lib/workshops-query.ts
```diff
@@ -199,16 +199,22 @@ export async function updateWorkshop(input: Module) {
...
-		fields: {
-			...currentWorkshop.fields,
-			...input.fields,
-			slug: workshopSlug,
-		},
-	})
+	return {
+		...updatedWorkshop,
+		resources: {},
+	}
```
vojtaholik, line 219: `onUpdate` form is expecting to return `ContentResource` type across multiple apps and so instead of changing that I ended up passing empty resources here. I don't think we need them anyway

PR #215, opened 2024-07-25
feat(pnjs): module purchase validation abilities

vojtaholik:
Fixes PNJS-3

image

gif

### DIFF
--- a/apps/pro-nextjs/src/ability/index.ts
+++ b/apps/pro-nextjs/src/ability/index.ts
@@ -8,16 +8,38 @@ import {
 } from '@casl/ability'
 import z from 'zod'
 
-import { userSchema } from '@coursebuilder/core/schemas'
+import { userSchema, type Purchase } from '@coursebuilder/core/schemas'
 import { ContentResourceResourceSchema } from '@coursebuilder/core/schemas/content-resource-schema'
 import { ContentResource } from '@coursebuilder/core/types'
+
+import {
+	hasAvailableSeats,
+	hasBulkPurchase,
+	hasChargesForPurchases,
+} from './purchase-validators'
 
 export const UserSchema = userSchema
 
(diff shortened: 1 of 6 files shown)


apps/pro-nextjs/src/ability/index.ts
```diff
@@ -108,65 +127,59 @@ export function defineRulesForPurchases(
...
+			}
+
+			if (purchase.status === 'Restricted' && purchase.country !== country) {
+				return { valid: false, reason: 'region_restricted' }
+			}
+
+			// if (purchase.status === 'Restricted' && module.type === 'bonus') {
+			// 	return { valid: false, reason: 'region_restricted' }
+			// }
+
```
vojtaholik, line 149: bonus module type doesn't exist yet so I'm leaving it out

apps/pro-nextjs/src/ability/index.ts
```diff
@@ -108,65 +127,59 @@ export function defineRulesForPurchases(
...
+			userHasPurchaseWithAccess.some(
+				(purchase) => purchase.reason === 'region_restricted',
+			)
+		) {
+			can('read', 'RegionRestriction')
+		}
+	}
+
+	if (hasChargesForPurchases(purchases)) {
+		can('read', 'Invoice')
```
vojtaholik, line 173: I've noticed that CB uses `read` instead of `view` even though both are supported. I think I like `view` more.

apps/pro-nextjs/src/ability/purchase-validators.test.ts
```diff
@@ -0,0 +1,97 @@
+import type { Coupon, Purchase } from '@coursebuilder/core/schemas'
```
vojtaholik, line 1: test file taken from skill stack

PR #217, opened 2024-07-25
feat(cb): handle stripe refund and dispute webhook events

zacjones93:
This adds a `PURCHASE_STATUS_UPDATED_EVENT` that we can send to inngest with a charge Id (identifier which is different than the `merchantChargeId` we store on a purchase) and status. For the current use-case, this sends `Refunded` and `Disputed` status's when we receive `charge.refunded`  or `charge.dispute.created` webhook events from stripe.

The result is a purchase with it's status updated:
image

gif

### DIFF
--- a/packages/adapter-drizzle/src/lib/mysql/index.ts
+++ b/packages/adapter-drizzle/src/lib/mysql/index.ts
@@ -1642,11 +1642,45 @@ export function mySqlDrizzleAdapter(
 		}): Promise<unknown> {
 			throw new Error('Method not implemented.')
 		},
-		updatePurchaseStatusForCharge(
+		async updatePurchaseStatusForCharge(
 			chargeId: string,
 			status: 'Valid' | 'Refunded' | 'Disputed' | 'Banned' | 'Restricted',
 		): Promise<Purchase | undefined> {
-			throw new Error('updatePurchaseStatusForCharge Method not implemented.')
+			const merchantChargeForPurchase =
+				await client.query.merchantCharge.findFirst({
+					where: eq(merchantCharge.identifier, chargeId),
+				})
+
+			const parsedMerchantChargeForPurchase = merchantChargeSchema.parse(
+				merchantChargeForPurchase,
+			)
+
+			if (!parsedMerchantChargeForPurchase)
+				throw new Error(`no-charge-found-for-purchase ${chargeId}`)
+
+			const purchase = await client.query.purchases.findFirst({
+				where: eq(
+					purchaseTable.merchantChargeId,
+					parsedMerchantChargeForPurchase.id,
+				),
+			})
+
+			const parsedPurchase = purchaseSchema.nullable().parse(purchase)
+
+			if (parsedPurchase) {
+				await client
+					.update(purchaseTable)
+					.set({ status: status })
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/pricing/process-stripe-webhook.ts
```diff
@@ -133,10 +224,22 @@ export async function processStripeWebhook(
 
 			break
 		case 'charge.refunded':
-			console.log('charge.refunded', { event })
+			await options.inngest.send({
+				name: PURCHASE_STATUS_UPDATED_EVENT,
+				data: {
+					stripeChargeId: event.data.object.id,
+					status: 'Refunded',
+				},
```
zacjones93, line 232: we send the whole stripe event for checkout session but for refunds and disputes it seems like we just need id and status.

packages/core/src/lib/pricing/process-stripe-webhook.ts
```diff
@@ -114,6 +115,96 @@ const exampleEvent = {
...
+const exampleRefundEvent = {
+	id: 'evt_3PfrkvCsSasMEpCd0eCwd9JH',
+	object: 'event',
+	api_version: '2023-10-16',
+	created: 1721927689,
+	data: {
+		object: {
+			id: 'ch_3PfrkvCsSasMEpCd0DEIwaHw',
+			object: 'charge',
+			amount: 29900,
```
zacjones93, line 127: I added this refund event for an example

packages/core/src/inngest/commerce/event-purchase-status-updated.ts
```diff
@@ -0,0 +1,55 @@
...
+	step,
+	db,
+	notificationProvider,
+	paymentProvider,
+}: CoreInngestFunctionInput) => {
+	return await step.run('update purchase status', async () => {
+		return db.updatePurchaseStatusForCharge(
+			event.data.stripeChargeId,
+			event.data.status,
+		)
```
zacjones93, line 47: the db provider has a `updatePurchaseStatusForCharge` function that needed to be implemented.

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -1642,11 +1642,45 @@ export function mySqlDrizzleAdapter(
...
+					.where(eq(purchaseTable.id, parsedPurchase.id))
+
+				const newPurchase = await client.query.purchases.findFirst({
+					where: eq(purchaseTable.id, parsedPurchase.id),
+				})
+
+				return purchaseSchema.optional().parse(newPurchase)
+			} else {
+				throw new Error(`no-purchase-found-for-charge ${chargeId}`)
+			}
```
zacjones93, line 1683: needed to find the `merchantCharge` then use the `id` to find the `purchase` and update `status`

packages/core/src/lib/pricing/process-stripe-webhook.ts
joelhooks: just sending the id is arguably better practice

"thin vs fat events"

depends, tradeoffs

joelhooks (approved):

PR #219, opened 2024-07-26
feat: ability to remove and reupload a video

joelhooks:
season 14 episode 6 GIF

### DIFF
--- a/packages/core/src/inngest/video-processing/functions/video-uploaded.ts
+++ b/packages/core/src/inngest/video-processing/functions/video-uploaded.ts
@@ -69,6 +69,22 @@ const videoUploadedHandler: CoreInngestHandler = async ({
 	}
 
 	if (videoResource && videoResource.id) {
+		const parentResource = await step.run('get parent resource', async () => {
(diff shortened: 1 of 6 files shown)


packages/core/src/inngest/video-processing/functions/video-uploaded.ts
```diff
@@ -69,6 +69,22 @@ const videoUploadedHandler: CoreInngestHandler = async ({
...
+		for (const resource of parentResource?.resources || []) {
+			if (resource.resource.type === 'videoResource') {
+				await step.run('detach existing video resource', async () => {
+					await db.removeResourceFromResource({
+						childResourceId: resource.resource.id,
+						parentResourceId: event.data.parentResourceId,
+					})
+				})
+			}
+		}
```
joelhooks, line 86: only going to allow 1 attached video resource, but we won't delete the previously uploaded resource so it could potentially still be used if wanted (ie this doesn't delete it, just disassociation)

Creeland (approved):

Creeland:
I agree with keeping the record of the original in case it's needed

PR #235, opened 2024-07-30
feat(pnjs): load progress with less queries

joelhooks:
perf

### DIFF
--- a/packages/adapter-drizzle/src/lib/mysql/index.ts
+++ b/packages/adapter-drizzle/src/lib/mysql/index.ts
@@ -995,176 +995,71 @@ export function mySqlDrizzleAdapter(
 		async getModuleProgressForUser(
 			userIdOrEmail: string,
 			moduleIdOrSlug: string,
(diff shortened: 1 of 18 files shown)


packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -995,193 +995,111 @@ export function mySqlDrizzleAdapter(
...
+        WHERE
+            (workshop.id = ${moduleIdOrSlug} OR workshop.fields->>'$.slug' = ${moduleIdOrSlug})
+            AND (
+                (cr.id = cr1.id AND cr1.type IN ('lesson', 'exercise'))
+                OR (cr.id = cr2.id AND cr2.type IN ('lesson', 'exercise'))
+            )
+        ORDER BY
+            COALESCE(crr1.position, 0),
+            COALESCE(crr2.position, 0);
+    `)
```
joelhooks, line 1051: less querying and data to get what we want

PR #243, opened 2024-08-01
feat(jsv): replace convertkit signup form with coursebuilder signup form

joelhooks:
sign me up


packages/core/src/lib/actions/index.ts
```diff
@@ -113,6 +122,28 @@ export async function subscribeToList(
...
+						email: user.email,
+						type: 'signup',
+						callbackUrl: `${options.baseUrl}/confirmed`,
+						emailProvider,
+						authOptions: options.authConfig,
+						adapter: options.adapter,
+						baseUrl: options.baseUrl,
+					})
+				}
+			}
```
joelhooks, line 146: subscribe to list and email verification are two separate steps for now and needs to be considered in more depth. This sends a convertkit style "click to confirm" email that sets the `emailVerified` property on the user to the date. This type of "double opt-in" is critical long term. We don't want to send broadcast/marketing emails to unverified recipients in general.

packages/core/src/lib/send-verification-request.ts
```diff
@@ -79,6 +82,14 @@ export const sendVerificationRequest = async (
...
 			} (${host})`
 			break
+		case 'signup':
+			subject = `Welcome to ${
+				process.env.NEXT_PUBLIC_PRODUCT_NAME ||
+				process.env.NEXT_PUBLIC_SITE_TITLE
+			} (${host})`
+			html = signUpHtml
+			text = signUpText
+			break
```
joelhooks, line 92: this was already a case we just weren't using it anywhere so here it's being used as a confirmation step

PR #245, opened 2024-08-03
fix(cb): properly identify bulk purchases

joelhooks:
bulk

### DIFF
--- a/packages/adapter-drizzle/src/lib/mysql/index.ts
+++ b/packages/adapter-drizzle/src/lib/mysql/index.ts
@@ -307,7 +307,7 @@ export function mySqlDrizzleAdapter(
 			if (coupon && couponValidation.isRedeemable) {
 				// if the Coupon is the Bulk Coupon of a Bulk Purchase,
 				// then a bulk coupon is being redeemed
-				const bulkCouponRedemption = Boolean(coupon.bulkPurchase?.bulkCouponId)
+				const bulkCouponRedemption = Boolean(coupon.maxUses > 1)
 
 				const { user } = await adapter.findOrCreateUser(email)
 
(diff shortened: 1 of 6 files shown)


packages/adapter-drizzle/src/lib/mysql/schemas/commerce/coupon.ts
```diff
@@ -73,10 +73,8 @@ export function getCouponRelationsSchema(mysqlTable: MySqlTableFn) {
 			references: [getProductSchema(mysqlTable).id],
 			relationName: 'product',
 		}),
-		bulkPurchase: one(purchase, {
-			fields: [coupon.bulkPurchaseId],
-			references: [purchase.id],
-			relationName: 'bulkPurchase',
+		bulkPurchases: many(purchase, {
+			relationName: 'bulkCoupon',
```
joelhooks, line 77: we have many bulk purchases for a single coupon since we add to them over time if they buy more stuff

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -307,7 +307,7 @@ export function mySqlDrizzleAdapter(
 			if (coupon && couponValidation.isRedeemable) {
 				// if the Coupon is the Bulk Coupon of a Bulk Purchase,
 				// then a bulk coupon is being redeemed
-				const bulkCouponRedemption = Boolean(coupon.bulkPurchase?.bulkCouponId)
+				const bulkCouponRedemption = Boolean(coupon.maxUses > 1)
```
joelhooks, line 310: you could check if the length of the bulkPurchases array was greater than 0 also but if it has max uses greater than 1 it's bulk coupon - this  logic is what was causing it to not be recorded since we weren't populating bulkPurchaseId. that's correct, we shouldn't since it is a `many` relationship, but it's confusing.

TODO: remove the bulkPurchaseId column from the coupon table since it is further confusing something that is confusing already 😭

PR #257, opened 2024-08-06
fix(pnjs): workshop lesson navigation sorting

joelhooks:
sorting


apps/pro-nextjs/src/lib/workshops-query.ts
```diff
@@ -23,49 +25,58 @@ import z from 'zod'
 import { productSchema } from '@coursebuilder/core/schemas'
 import { ContentResource } from '@coursebuilder/core/types'
 
-export async function getWorkshopNavigation(
-	moduleSlugOrId: string,
-): Promise<WorkshopNavigation | null> {
+async function getAllWorkshopLessonsWithSectionInfo(moduleSlugOrId: string) {
```
joelhooks, line 28: this query returns a row for every lesson in a given module assuming:

* top-level lessons aren't in a section
* sections can only contain lessons, not more sections

each row contains the lesson info, and if it is in a section, the section info as well. There are two position columns in the rows. One is for the top level sort and the other is for lesson in a section

apps/pro-nextjs/src/lib/workshops-query.ts
```diff
@@ -135,37 +152,38 @@ ORDER BY
...
-				title: item.lesson_title || item.section_or_lesson_title!,
-				position: item.lesson_position || item.section_or_lesson_position!,
+			const newLesson: NavigationLesson = NavigationLessonSchema.parse({
+				id: item.lesson_id,
+				slug: item.lesson_slug,
+				title: item.lesson_title,
+				position: item.lesson_position,
 				type: 'lesson',
-			}
+			})
```
joelhooks, line 161: top-level lessons

apps/pro-nextjs/src/lib/workshops-query.ts
```diff
@@ -135,37 +152,38 @@ ORDER BY
...
+					title: item.section_title,
+					position: item.section_position,
 					type: 'section',
 					lessons: [],
-				}
-				sectionsMap.set(item.section_or_lesson_id, newSection)
+				})
+				sectionsMap.set(item.section_id, newSection)
 				resources.push(newSection)
 			}
```
joelhooks, line 176: create a section container for the nav UI

apps/pro-nextjs/src/lib/workshops-query.ts
```diff
@@ -135,37 +152,38 @@ ORDER BY
...
+			if (item.lesson_id) {
+				const newLesson: NavigationLesson = NavigationLessonSchema.parse({
+					id: item.lesson_id,
+					slug: item.lesson_slug,
+					title: item.lesson_title,
+					position: item.lesson_position,
+					type: 'lesson',
+				})
+				sectionsMap.get(item.section_id)?.lessons.push(newLesson)
+			}
```
joelhooks, line 186: add lesson to section it belongs to

PR #269, opened 2024-08-16
feat(pnjs): wire up discord

joelhooks:


apps/pro-nextjs/src/inngest/functions/discord/discord-account-linked.ts
```diff
@@ -0,0 +1,117 @@
...
+								roles: Array.from(
+									new Set([...discordMember.roles, env.DISCORD_MEMBER_ROLE_ID]),
+								),
+							}),
+							headers: {
+								'Content-Type': 'application/json',
+							},
+						},
+					)
+				}
```
joelhooks, line 80: we add an arbitrary role here, next step will be to check for an active purchase and give them a special role as a "member"

we will also check if the user account is connected to discord at the time of purchase and apply that role there as well

THEN if they aren't we will present them with a "link to discord" opportunity clearly

apps/pro-nextjs/src/server/auth.ts
```diff
@@ -51,6 +53,13 @@ export const authOptions: NextAuthConfig = {
 		createUser: async ({ user }) => {
 			await inngest.send({ name: USER_CREATED_EVENT, user, data: {} })
 		},
+		linkAccount: async ({ user, account, profile }) => {
+			await inngest.send({
+				name: OAUTH_PROVIDER_ACCOUNT_LINKED_EVENT,
+				data: { account, profile },
+				user,
+			})
+		},
```
joelhooks, line 62: standard Auth.js event that we forward into inngest

apps/pro-nextjs/src/server/auth.ts
```diff
@@ -107,6 +116,20 @@ export const authOptions: NextAuthConfig = {
...
 				]
 			: []),
+		...(env.NEXT_PUBLIC_DISCORD_CLIENT_ID && env.DISCORD_CLIENT_SECRET
+			? [
+					DiscordProvider({
+						clientId: env.NEXT_PUBLIC_DISCORD_CLIENT_ID,
+						clientSecret: env.DISCORD_CLIENT_SECRET,
+						allowDangerousEmailAccountLinking: true,
+						authorization:
+							'https://discord.com/api/oauth2/authorize?scope=identify+email+guilds.join',
```
joelhooks, line 126: sending in an override so that we can expand the scope to what we need for integration

PR #275, opened 2024-08-26
feat(vbd): upload and download lesson assets

zacjones93:
https://www.loom.com/share/9dd3bd676bd54dceadb6a227c510f23a

gif


apps/value-based-design/src/trpc/api/routers/imageResource.ts
```diff
@@ -1,8 +1,78 @@
...
+			})
+		}),
+	download: publicProcedure
+		.input(
+			z.object({
+				resourceId: z.string(),
+				lessonTitle: z.string(),
+			}),
+		)
+		.mutation(async ({ input }) => {
```
joelhooks, line 37: this could probably just be a server action? most trpc mutations could be

PR #284, opened 2024-09-01
feat: moving product queries to adapter

joelhooks:
requires a payment provider as well so cleaning up the stripe stuff


apps/astro-party/src/coursebuilder/stripe-provider.ts
```diff
@@ -1,11 +1,15 @@
...
 
 export const stripeProvider = StripeProvider({
 	errorRedirectUrl: `${env.COURSEBUILDER_URL}`,
 	baseSuccessUrl: `${env.COURSEBUILDER_URL}`,
 	cancelUrl: `${env.COURSEBUILDER_URL}`,
-	paymentsAdapter: new StripePaymentAdapter(),
+	paymentsAdapter: new StripePaymentAdapter({
+		stripeToken: env.STRIPE_SECRET_TOKEN,
+		stripeWebhookSecret: env.STRIPE_WEBHOOK_SECRET,
+	}),
```
joelhooks, line 14: the payments adapter creates a stripe instance internally now instead of passing one in. i like this **less** technically because now we can't control the stripe instance if we wanted to mock or test or isolate

taking this a step further would be moving the creation of the PaymentAdapter into the PaymentProvider so the user app is passing in keys and the "adapter" idea is hidden 

still thinking about it

apps/astro-party/src/db/index.ts
```diff
@@ -14,4 +15,8 @@ export const db = drizzle(
 	{ schema },
 )
 
-export const courseBuilderAdapter = DrizzleAdapter(db, mysqlTable)
+export const courseBuilderAdapter = DrizzleAdapter(
+	db,
+	mysqlTable,
+	stripeProvider,
```
joelhooks, line 21: migrating functionality from `/lip/products-query` requires integration with payments to handle things like `createProduct` and other CRUD

apps/pro-nextjs/src/app/(commerce)/products/page.tsx
```diff
@@ -14,10 +13,7 @@ import {
 
 export default async function EventIndexPage() {
 	const { ability } = await getServerAuthSession()
-	const products = await db.query.products.findMany({
-		where: (products, { eq, and }) =>
-			and(eq(products.status, 1), eq(products.type, 'self-paced')),
-	})
```
joelhooks, line 20: moving all of the db calls into the lib is step 1, then to the adapter for distribution

moving stuff towards an adapter that is generic interface (`CourseBuilderAdapter`)

with `CourseBuilderAdapter` we are able to simplify and reduce the cognitive load of building apps

```typescript
const products = await db.query.products.findMany({
		where: (products, { eq, and }) =>
			and(eq(products.status, 1), eq(products.type, 'self-paced')),
	})
```

versus 

```typescript
const products = await courseBuilderAdapter.getActiveSelfPacedProducts()
```

we still have the freedom to write raw SQL or use the ORM as we please, but the simplest path is simple.

packages/adapter-drizzle/src/index.ts
```diff
@@ -11,14 +17,19 @@ import { type SqlFlavorOptions, type TableFn } from './lib/utils.js'
 
 export function DrizzleAdapter<SqlFlavor extends SqlFlavorOptions>(
 	db: SqlFlavor,
-	table?: TableFn<SqlFlavor>,
+	table: TableFn<SqlFlavor>,
+	paymentProvider: PaymentsProviderConfig = MockStripeProvider,
```
joelhooks, line 21: the `MockStripeProvider` is basically a `noop`

this can be better

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -1222,12 +1227,327 @@ export function mySqlDrizzleAdapter(
...
+						JSON.stringify(parsedProduct.error),
+						JSON.stringify(productData),
+					)
+					return null
+				}
+				return parsedProduct.data
+			} catch (e) {
+				console.log('getProduct error', e)
+				return null
+			}
```
joelhooks, line 1550: all of these methods have a good chunk of logic and now it isn't repeated 5~ times across apps lol

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -1222,12 +1227,327 @@ export function mySqlDrizzleAdapter(
...
+
+			const stripePrice = await paymentProvider.createPrice({
+				product: stripeProduct.id,
+				unit_amount: Math.floor(Number(input.price) * 100),
+				currency: 'usd',
+				nickname: input.name,
+				metadata: {
+					slug: product?.fields?.slug || null,
+				},
+			})
```
joelhooks, line 1477: here is where the payments provider is useful since it is an interface we aren't exposing Stripe directly and instead exposing a "concept"

it's stripe specific but any payment provider is likely to have the CONCEPT of a price so our api can be like Stripe and that's **totally fine**

Stripe knows whats up

joelhooks: one red flag here would be adding MORE providers like this, but imo we shouldn't need to

commerce is "special" and `CourseBuilderAdapter` needs to be able to communicate with that specific external database

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -1222,12 +1227,327 @@ export function mySqlDrizzleAdapter(
...
+				status: 1,
+			})
+
+			// TODO: handle upgrades
+
+			return product
+		},
+		async getProduct(
+			productSlugOrId?: string,
+			withResources: boolean = true,
```
joelhooks, line 1505: made this an option because sometimes you don't want resources. probably need an "option bag" to control price as well?

packages/adapter-drizzle/test/mysql/schema.ts
```diff
@@ -55,8 +59,14 @@ export const {
...
 	contributionTypesRelations,
 	resourceProgress,
+	contentResourceProduct,
+	contentResourceProductRelations,
 	upgradableProducts,
 	upgradableProductsRelations,
+	comments,
+	commentsRelations,
+	userPrefs,
+	userPrefsRelations,
```
joelhooks, line 69: these were incomplete for commerce stuff and was throwing errors

packages/core/src/providers/stripe.ts
```diff
@@ -64,9 +64,167 @@ export default function StripeProvider(
...
+	) {
+		await this.stripe.prices.update(priceId, price)
+	}
+	async createPrice(price: Stripe.PriceCreateParams) {
+		return this.stripe.prices.create(price)
+	}
+	async createProduct(product: Stripe.ProductCreateParams) {
+		return this.stripe.products.create(product)
 	}
 }
```
joelhooks, line 207: this adapter was in every project, so now it's in core as it should be, just one copy of the code

packages/core/src/types.ts
```diff
@@ -140,6 +156,22 @@ export interface PaymentsAdapter {
...
+	updatePrice<TPriceUpdate = Stripe.Price>(
+		priceId: string,
+		priceUpdate: Partial<TPriceUpdate>,
+	): Promise<void>
+	createPrice(
+		price: Stripe.PriceCreateParams,
+	): Promise<Stripe.Response<Stripe.Price>>
+	createProduct(
+		product: Stripe.ProductCreateParams,
+	): Promise<Stripe.Response<Stripe.Product>>
```
joelhooks, line 174: this is annoying to have repeated and they should be the same anyway probably

packages/adapter-drizzle/src/lib/mysql/index.ts
```diff
@@ -1222,12 +1227,327 @@ export function mySqlDrizzleAdapter(
 				}),
 			)
 		},
-		async getProduct(productId: string): Promise<Product | null> {
-			return productSchema.nullable().parse(
-				await client.query.products.findFirst({
-					where: eq(products.id, productId),
+		async archiveProduct(productId) {
```
joelhooks, line 1230: these can get integration tests now 🏆

apps/astro-party/package.json
```diff
@@ -49,7 +49,6 @@
 		"@coursebuilder/core": "0.2.2",
 		"@coursebuilder/email-templates": "^1.0.2",
 		"@coursebuilder/next": "workspace:^",
-		"@coursebuilder/react-rsc": "workspace:^",
```
joelhooks, line 52: this package isn't needed and the single component moved over to `ui`

apps/astro-party/src/app/(commerce)/products/[slug]/_components/product-pricing.tsx
```diff
@@ -6,11 +6,8 @@ import { env } from '@/env.mjs'
 import { pricingClassNames } from '@/styles/commerce'
 
 import { PriceCheckProvider } from '@coursebuilder/commerce-next/pricing/pricing-check-context'
-import {
-	PricingData,
-	PricingWidget,
-} from '@coursebuilder/commerce-next/pricing/pricing-widget'
-import { CommerceProps } from '@coursebuilder/commerce-next/utils/commerce-props'
+import { PricingWidget } from '@coursebuilder/commerce-next/pricing/pricing-widget'
+import { CommerceProps, PricingData } from '@coursebuilder/core/types'
```
Comment (joelhooks) [inline apps/astro-party/src/app/(commerce)/products/[slug]/_components/product-pricing.tsx:10]: moving types out of other packages and into `core` around pricing since `commerce-next` is poorly named to begin with

still thinking about package naming changes

apps/astro-party/src/app/(content)/_components/video-player-overlay.tsx
```diff
@@ -18,14 +18,14 @@ import pluralize from 'pluralize'
 import { useFormStatus } from 'react-dom'
 
 import InviteTeam from '@coursebuilder/commerce-next/team/invite-team'
-import { buildStripeCheckoutPath } from '@coursebuilder/commerce-next/utils/build-stripe-checkout-path'
-import { formatUsd } from '@coursebuilder/commerce-next/utils/format-usd'
+import { buildStripeCheckoutPath } from '@coursebuilder/core/pricing/build-stripe-checkout-path'
```
joelhooks, line 21: these utils make sense in core too, maybe in the stripe provider?

apps/astro-party/src/app/(content)/workshops/_components/workshop-pricing-server.tsx
```diff
@@ -38,6 +39,10 @@ export async function WorkshopPricing({
 			productId: product.id,
 		})
 
+		const countryCode =
+			headers().get('x-vercel-ip-country') ||
+			process.env.DEFAULT_COUNTRY ||
+			'US'
```
joelhooks, line 45: we don't need headers beyond just getting the country code so we can do that app side and pass it in so we don't have a next dep to deal with

vojtaholik: nice catch!

vojtaholik (approved):
this is great! 🚀

PR #293, opened 2024-09-13
feat(vbd): Set lesson to free with isFreeToView field

zacjones93:
A pattern we use a lot is to set the first lesson in a course (or sometimes sections) to free. We've had this hard-coded to the first lesson in the lesson list which doesn't allow for much flexibility. This converts that functionality to check for a `isFreeToView` field value on lesson and displays the lesson if it's true

image

lesson page

gif

### DIFF
--- a/apps/value-based-design/src/ability/index.ts
+++ b/apps/value-based-design/src/ability/index.ts
@@ -232,24 +232,7 @@ const isFreelyVisible = ({
 	lesson,
 	isSolution,
 }: ViewerAbilityInput) => {
-	const hasId = lesson && 'id' in lesson
-
-	// return false if it is a 'Solution'
-	if (isSolution || lesson?.type === 'solution' || !hasId) {
-		return false
-	}
-
-	const lessons = z
-		.array(z.object({ resourceId: z.string() }))
-		.parse(section ? section.resources : module?.resources || [])
-
-	const isFirstLesson =
-		(lesson?.type === 'exercise' ||
-			lesson?.type === 'explainer' ||
-			lesson?.type === 'lesson') &&
-		lesson.id === lessons?.[0]?.resourceId
-
-	return isFirstLesson && lesson && !isSolution
+	return lesson?.fields?.isFreeToView
 }
 
 /**
--- a/apps/value-based-design/src/lib/workshops-query.ts
+++ b/apps/value-based-design/src/lib/workshops-query.ts
@@ -50,7 +50,12 @@ async function getAllWorkshopLessonsWithSectionInfo(
     combined.lesson_id,
     combined.lesson_slug,
     combined.lesson_title,
-    combined.lesson_position
+    combined.lesson_position,
+    CASE
+        WHEN combined.isFreeToView = 'true' THEN TRUE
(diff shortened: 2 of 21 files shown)


apps/value-based-design/src/lib/workshops-query.ts
```diff
@@ -50,7 +50,12 @@ async function getAllWorkshopLessonsWithSectionInfo(
     combined.lesson_id,
     combined.lesson_slug,
     combined.lesson_title,
-    combined.lesson_position
+    combined.lesson_position,
+    CASE
+        WHEN combined.isFreeToView = 'true' THEN TRUE
+        WHEN combined.isFreeToView = 'false' THEN FALSE
+        ELSE NULL
+    END AS isFreeToView
```
zacjones93, line 58: This returns a `1` or `0` if the field is true or false otherwise the query would return a string of 'true' or 'false' which didn't seem great

apps/value-based-design/src/lib/workshops-query.ts
```diff
@@ -168,6 +177,7 @@ export async function getWorkshopNavigation(
 				slug: item.lesson_slug,
 				title: item.lesson_title,
 				position: item.lesson_position,
+				isFreeToView: Boolean(item.isFreeToView),
 				type: 'lesson',
 			})
```
zacjones93, line 182: here we return `Boolean(item.isFreeToView)` to coerce that `1` or `0` into `true` or `false`

apps/value-based-design/src/lib/workshops.ts
```diff
@@ -14,6 +14,7 @@ export const NavigationResultSchema = z.object({
 	lesson_slug: z.string().nullable(),
 	lesson_title: z.string().nullable(),
 	lesson_position: z.number().nullable(),
+	isFreeToView: z.number().optional().nullable(),
```
zacjones93, line 17: navigation result is a number..

apps/value-based-design/src/lib/workshops.ts
```diff
@@ -23,6 +24,7 @@ export const NavigationLessonSchema = z.object({
 	slug: z.string(),
 	title: z.string(),
 	position: z.number(),
+	isFreeToView: z.boolean().optional().nullable(),
```
zacjones93, line 27: NavigationLessonSchema ends up with `isFreeToView` as a boolean

apps/value-based-design/src/ability/index.ts
```diff
@@ -232,24 +232,7 @@ const isFreelyVisible = ({
...
-		.parse(section ? section.resources : module?.resources || [])
-
-	const isFirstLesson =
-		(lesson?.type === 'exercise' ||
-			lesson?.type === 'explainer' ||
-			lesson?.type === 'lesson') &&
-		lesson.id === lessons?.[0]?.resourceId
-
-	return isFirstLesson && lesson && !isSolution
+	return lesson?.fields?.isFreeToView
```
zacjones93, line 235: `isFreelyVisible` now just needs to check if the lesson's `isFreeToView` field value is set to true

joelhooks (approved):

PR #299, opened 2024-09-27
feat(ui): module navigation primitives

vojtaholik:
spike for module navigation list components that reside in the UI package with no Next.js dependencies. I began by abstracting leaf components with accessibility features, while the final component is still assembled at the app level. this feels like a solid starting point and allows us to consider separation of concerns more effectively.

gif


packages/ui/module-navigation/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+	type: z.literal('section'),
+	lessons: z.array(NavigationLessonSchema),
+})
+
+export const NavigationResourceSchema = z.discriminatedUnion('type', [
+	NavigationSectionSchema,
+	NavigationLessonSchema,
+])
+
+type NavigationResource = z.infer<typeof NavigationResourceSchema>
```
vojtaholik, line 37: todo: figure out where these types should live

apps/pro-nextjs/src/app/_components/module-navigation-client.tsx
```diff
@@ -0,0 +1,160 @@
...
+		workshopNavigation,
+		currentResourceSlug,
+	)
+	const { data: abilityRules, status: abilityStatus } =
+		api.ability.getCurrentAbilityRules.useQuery({
+			moduleId: workshopNavigation?.id,
+		})
+	const ability = createAppAbility(abilityRules || [])
+
+	return (
```
vojtaholik, line 34: currently, the final component is assembled at the app level, but I’m considering how it could be further abstracted, especially from a data perspective. 🤔

joelhooks (approved):

PR #307, opened 2024-10-09
fix(egh): safely get videoResource for current post

zacjones93:
Some posts (articles) don't have a video so when we load the page on the server we should opt out of fetching with `getVideoResource` 

The `EditPostForm` component already handles `videoResourceId` and `videoResourceLoader` being `null` so this is a simple fix here

currently blowing up '[Usecases for TypeScript Brand Types](https://builder.egghead.io/usecases-for-type-script-brand-types~nzix2/edit)' edit page

image

gif


apps/egghead/src/app/(content)/posts/[slug]/edit/page.tsx
```diff
@@ -35,9 +35,10 @@ export default async function PostPage({
 				return resource.type === 'videoResource'
 			}) || null
 
-	const videoResourceLoader = courseBuilderAdapter.getVideoResource(
-		videoResource.id,
-	)
+	const videoResourceLoader = videoResource
+		? courseBuilderAdapter.getVideoResource(videoResource?.id)
```
Comment (joelhooks) [inline apps/egghead/src/app/(content)/posts/[slug]/edit/page.tsx:39]: i think this should be handled in getVideoResource so that if it is undefined/null it returns null from there

joelhooks (approved):

apps/egghead/src/app/(content)/posts/[slug]/edit/page.tsx
Comment (zacjones93) [reply apps/egghead/src/app/(content)/posts/[slug]/edit/page.tsx:39]: https://linear.app/skillrecordings/issue/CB-20/coursebuilderadaptergetvideoresource-return-null-when-if-videoresource

PR #314, opened 2024-11-01
feat: This updates Radix, TipTap, and the React calendar that we use in coupons.

joelhooks:
This is an initial pass to update Radix, Tip Tap, and the React Calendar in the process of getting us to Next.js 15 and the React 19 upgrade.


apps/ai-hero/package.json
```diff
@@ -99,11 +99,11 @@
 		"@tanstack/react-query": "5.54.1",
 		"@tanstack/react-query-devtools": "5.54.1",
 		"@tanstack/react-table": "^8.17.3",
-		"@tiptap/extension-highlight": "^2.6.4",
-		"@tiptap/extension-link": "^2.6.4",
-		"@tiptap/extension-typography": "^2.6.4",
-		"@tiptap/react": "^2.6.4",
-		"@tiptap/starter-kit": "^2.6.4",
+		"@tiptap/extension-highlight": "2.9.1",
```
joelhooks, line 102: I don't actually know what the issue was here, but somehow 2.9.1 was actually installed and I was having a conflict, so I just set it to that. We hardcoded it to 9.1 to maybe avoid that in the future.

apps/pro-nextjs/src/app/admin/coupons/_components/coupon-generator-form.tsx
```diff
@@ -206,7 +206,6 @@ const CouponGeneratorForm = ({
 											disabled={(date) =>
 												date < new Date() || date < new Date('1900-01-01')
 											}
-											initialFocus
```
joelhooks, line 209: It looks like they removed this from the api.

packages/ui/primitives/calendar.tsx
```diff
@@ -52,8 +52,12 @@ function Calendar({
...
 			}}
 			components={{
-				IconLeft: ({ ...props }) => <ChevronLeft className="h-4 w-4" />,
-				IconRight: ({ ...props }) => <ChevronRight className="h-4 w-4" />,
+				Chevron: (props) => {
+					if (props.orientation === 'left') {
+						return <ChevronLeft className="h-4 w-4" />
+					}
+					return <ChevronRight className="h-4 w-4" />
+				},
```
joelhooks, line 60: This was a documented change in the way we pass properties into here.

PR #319, opened 2024-11-08
feat(egh): add slack notification for post create

zacjones93:
gif


apps/egghead/src/lib/posts-query.ts
```diff
@@ -211,6 +214,13 @@ export async function createPost(input: NewPost) {
 	})
 
 	if (post) {
+		await inngest.send({
+			name: POST_CREATED_EVENT,
+			data: {
+				post: await post,
+			},
+		})
+
```
zacjones93, line 223: The one thing I'm not sure about is awaiting the post here in `post-query.ts`, when I sent over the post as a promise the data was undefined 🤔

apps/egghead/src/lib/posts.ts
```diff
@@ -56,7 +56,7 @@ export const PostSchema = ContentResourceSchema.merge(
 			github: z.string().nullish(),
 			gitpod: z.string().nullish(),
 		}),
-		tags: z.array(z.any()),
+		tags: z.array(z.any()).nullish(),
```
zacjones93, line 59: noticed while deleting test posts that there was an issue with the zod schema for posts with no tags

apps/egghead/src/inngest/functions/notify-slack-for-post.ts
```diff
@@ -0,0 +1,47 @@
...
+import { POST_CREATED_EVENT } from '../events/post-created'
+
+export const notifySlack = inngest.createFunction(
+	{
+		id: `notify-slack`,
+		name: `Notify Slack`,
+	},
+	{
+		event: POST_CREATED_EVENT,
+	},
```
zacjones93, line 15: This should be batched in inngest

https://www.inngest.com/docs/guides/batching

apps/egghead/src/lib/posts-query.ts
joelhooks: `writeNewPostToDatabase` should probably be awaited above to be more "normal"

apps/egghead/src/lib/posts.ts
joelhooks: better to have a default empty array generally

joelhooks (approved):

PR #320, opened 2024-11-12
chore(cb): orgs

joelhooks:
This is an example of a large data-oriented migration where we want to move from a user-centric structure in terms of how accounts are managed into an organization-level structure. What that means is we can't go through and just wipe users. We need to add organizations, do the migration, and then come back afterwards after we maybe have them running in parallel for a little while and then pull users out so we only have one way to think about and identify ownership within the application.

It's going to give us a lot of flexibility, particularly when we start talking about team accounts first, subscription properties, that sort of thing. But it's just a better way to kind of lay it out to plan for the future, using some tried and true techniques.


apps/ai-hero/package.json
```diff
@@ -5,7 +5,7 @@
 	"type": "module",
 	"scripts": {
 		"dev": "concurrently --kill-others-on-fail \"npm:dev:*\"",
-		"dev:next": "next dev --turbo",
```
joelhooks, line 8: Turbos pretty much just given me trouble.

apps/ai-hero/src/db/schema.ts
```diff
@@ -60,4 +60,14 @@ export const {
...
+	organization,
+	organizationRelations,
+	organizationMemberships,
+	organizationMembershipRelations,
+	organizationMembershipRoles,
+	organizationMembershipRolesRelations,
+	merchantSubscription,
+	merchantSubscriptionRelations,
+	subscription,
+	subscriptionRelations,
```
joelhooks, line 72: This is the full stack of what's been added to this PR, mostly around organization. An organization has members, so there is basically a join table between the organization and the user called organization membership. Subscriptions and related items are also included here. They are owned by the organization, not by a member.

One of the key points here is that even a single user, like a team, is going to have an organization, probably a personal organization or whatever you would call it. This approach ensures that purchases will be consistent, whether it's a single person purchasing just for themselves or organizations that are teams.

packages/adapter-drizzle/src/lib/mysql/schemas/commerce/purchase.ts
```diff
@@ -22,6 +24,10 @@ export function getPurchaseSchema(mysqlTable: MySqlTableFn) {
 		{
 			id: varchar('id', { length: 191 }).notNull(),
 			userId: varchar('userId', { length: 191 }),
+			purchasedByorganizationMembershipId: varchar('organizationMembershipId', {
+				length: 191,
+			}),
+			organizationId: varchar('organizationId', { length: 191 }),
```
joelhooks, line 30: Technically speaking, I think organization ID is redundant here because we have the per purchase by organization membership ID, which is the join table that goes between them. Then you would have the organization through that. So we probably don't need to have the organization here.

packages/adapter-drizzle/src/lib/mysql/schemas/commerce/subscription.ts
```diff
@@ -0,0 +1,71 @@
...
+import { getMerchantSessionSchema } from './merchant-session.js'
+import { getMerchantSubscriptionSchema } from './merchant-subscription.js'
+import { getProductSchema } from './product.js'
+
+export function getSubscriptionSchema(mysqlTable: MySqlTableFn) {
+	return mysqlTable(
+		'Subscription',
+		{
+			id: varchar('id', { length: 191 }).notNull(),
+			organizationId: varchar('organizationId', { length: 191 }),
```
joelhooks, line 24: Subscriptions and purchases are both owned by an organization, not by individual organization members. There might be members who have specific roles that would allow them control over various subscriptions and purchases, etc., but they are owned by the organization.

packages/adapter-drizzle/src/lib/mysql/schemas/communication/comment.ts
```diff
@@ -10,13 +10,17 @@ import {
...
 
 export function getCommentsSchema(mysqlTable: MySqlTableFn) {
 	return mysqlTable(
 		'Comment',
 		{
 			id: varchar('id', { length: 191 }).notNull(),
 			userId: varchar('userId', { length: 255 }).notNull(),
+			organizationMembershipId: varchar('organizationMembershipId', {
+				length: 255,
+			}),
```
joelhooks, line 23: Any place you would typically have a user as the owner, objects like this, such as a comment, are owned by an organization member, not the user. This gives the organization some ownership of stuff that's created within its purview, like a comment. So if the user leaves the organization, you can choose then to leave it intact or remove it with the user.

packages/adapter-drizzle/src/lib/mysql/schemas/content/content-resource.ts
```diff
@@ -20,6 +21,13 @@ export function getContentResourceSchema(mysqlTable: MySqlTableFn) {
 		'ContentResource',
 		{
 			id: varchar('id', { length: 255 }).notNull().primaryKey(),
+			organizationId: varchar('organizationId', { length: 191 }),
+			createdByOrganizationMembershipId: varchar(
+				'createdByOrganizationMembershipId',
+				{
+					length: 191,
+				},
+			),
```
joelhooks, line 30: In this case, we again probably dropped the organization ID in favor of the created by organization membership ID, which would give us the joint table between this piece of content and the organization. This would provide that separation from created by ID, which would get removed here entirely. It'd be created by organization membership ID instead. Again, this separates the idea of the user object, which is a thing owned by a person, from the membership which belongs to an organization. So if for whatever reason a user left an organization, we could then have the choice to keep the content that was created for that organization or delete it. This approach gives us that ability to make a decision.

packages/adapter-drizzle/src/lib/mysql/schemas/communication/comment.ts
zacjones93: ahhh that make sense

packages/adapter-drizzle/src/lib/mysql/schemas/content/content-resource.ts
zacjones93: Nice to be able to make that decision

PR #321, opened 2024-11-12
feat(egh): Create an index of instructors

zacjones93:
This sets up an index of instructors under `/admin/instructors` that will be the place we initiate instructor invites as well as can look up instructor profile to edit 

image


apps/egghead/src/app/admin/layout.tsx
```diff
@@ -0,0 +1,53 @@
...
+					</SidebarMenu>
+				</SidebarContent>
+			</Sidebar>
+			<main className="w-full p-10">
+				<SidebarTrigger />
+				<div>{children}</div>
+			</main>
+		</SidebarProvider>
+	)
+}
```
zacjones93, line 51: added shadcn sidebar component for the admin layout which is nice

apps/egghead/src/app/admin/tags/page.tsx
```diff
@@ -8,7 +8,7 @@ import TagManagement from './tag-management-client-page'
 export default async function TagManagementPage() {
 	const { ability } = await getServerAuthSession()
 
-	if (!ability.can('manage', 'Content')) {
+	if (!ability.can('manage', 'all')) {
```
zacjones93, line 11: consistent ability checks

apps/egghead/src/lib/users.ts
```diff
@@ -9,6 +9,31 @@ import { and, eq } from 'drizzle-orm'
...
+		.innerJoin(userRoles, eq(users.id, userRoles.userId))
+		.innerJoin(roles, eq(userRoles.roleId, roles.id))
+		.where(eq(roles.name, 'contributor'))
+
+	if (!instructors) {
+		return []
+	}
+
+	return instructors
+}
```
zacjones93, line 36: adds get instructor query

joelhooks (approved):

PR #323, opened 2024-11-14
feat: Sync Lessons and Video Resources to Sanity

Creeland:
This PR introduces functionality to automatically sync lessons and their associated video resources from egghead to Sanity when new lessons are created.

## Changes

- Added new Inngest function `syncLessonToSanity` that creates a lesson document in Sanity when a new lesson is created in egghead
- Added new Inngest function `syncVideoResourceToSanity` that creates a video resource document in Sanity and associates it with the corresponding lesson
- Created new event type `EGGHEAD_LESSON_CREATED_EVENT` that triggers the sync process
- Fixed typo in function name: `crreateEggheadLesson` -> `createEggheadLesson`
- Updated `createEggheadLesson` to emit the lesson created event
- Added new Inngest functions to the config

## Notes

- The sync is one-way: egghead -> Sanity


apps/egghead/src/inngest/functions/sanity/sync-video-resource-to-sanity.ts
```diff
@@ -0,0 +1,153 @@
...
+			muxPlaybackId: z.string().optional(),
+		})
+		.optional(),
+	transcript: z
+		.object({
+			srt: z.string(),
+			text: z.string(),
+		})
+		.optional(),
+})
```
joelhooks, line 42: these could be moved to /libs/sanity-content.ts

apps/egghead/src/inngest/functions/sanity/sync-video-resource-to-sanity.ts
```diff
@@ -0,0 +1,153 @@
...
+							courseBuilderVideoResource?.id || '',
+						),
+					})
+
+				return await db.query.contentResource.findFirst({
+					where: eq(
+						contentResource.id,
+						lessonVideoContentResourceResource?.resourceOfId || '',
+					),
+				})
```
joelhooks, line 90: check out courseBuilderAdapter

apps/egghead/src/lib/sanity-content-query.ts
```diff
@@ -0,0 +1,152 @@
...
+	eggheadLessonId: number | null | undefined,
+) {
+	if (!eggheadLessonId) return
+
+	return sanityWriteClient.fetch(
+		`*[_type == "lesson" && railsLessonId == ${eggheadLessonId}][0]`,
+	)
+}
+
+export async function postSanityLesson(
```
joelhooks, line 85: "post" is probably not the correct term here

"create" "update" and "get" is what we generally use. Get is coincidental

PR #328, opened 2024-11-21
feat: sync video resource data

joelhooks:
basic initial sketch of a sync that keeps the hls url updated with data sources

this would be a good "hook" to update sanity as well


apps/egghead/src/inngest/functions/video-resource-created.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { inngest } from '../inngest.server'
+
+export const syncVideoResourceData = inngest.createFunction(
+	{
+		id: `egghead video resource created sync`,
+		name: 'egghead Video Resource Created Sync',
+	},
+	{
+		event: VIDEO_RESOURCE_CREATED_EVENT,
```
joelhooks, line 15: since we are already dispatching this event with every video it felt like a good spot to sync

apps/egghead/src/inngest/functions/video-resource-created.ts
```diff
@@ -0,0 +1,72 @@
...
+					return PostSchema.nullable().parse(
+						await courseBuilderAdapter.getContentResource(
+							resourceResource.resourceOfId,
+						),
+					)
+				}
+
+				return null
+			},
+		)
```
joelhooks, line 46: a new videoResource is generally speaking going to have a parent (post)

apps/egghead/src/inngest/functions/video-resource-created.ts
```diff
@@ -0,0 +1,72 @@
...
+
+		if (parentResource && videoResource) {
+			await step.run('update the parent resource', async () => {
+				const hlsUrl = `https://stream.mux.com/${videoResource.muxPlaybackId}.m3u8`
+				const duration = videoResource.duration ?? 0
+
+				// TODO: also write to Sanity?
```
joelhooks, line 53: we should update the sanity lesson too!

apps/egghead/src/inngest/functions/video-resource-created.ts
```diff
@@ -0,0 +1,72 @@
...
+                updated_at = NOW(),
+                current_video_hls_url = $2
+              WHERE id = $3`,
+						[
+							Math.floor(duration),
+							hlsUrl,
+							parentResource.fields.eggheadLessonId,
+						],
+					)
+				}
```
joelhooks, line 68: this needs to be checked and tested, but if `current_video_hls_url` is set the lesson object in the egghead database should produce the `hls_url` as expected so presentation can occur normally using the authentication bits from the rails api for the given lesson

apps/egghead/src/lib/egghead.ts
```diff
@@ -77,11 +77,12 @@ export async function createEggheadLesson(input: {
...
 	instructorId: string | number
+	hlsUrl?: string
 }) {
-	const { title, slug, guid, instructorId } = input
+	const { title, slug, guid, instructorId, hlsUrl = null } = input
 	const eggheadLessonResult = await eggheadPgQuery(
 		`INSERT INTO lessons (title, instructor_id, slug, resource_type, state,
-			created_at, updated_at, visibility_state, guid)
+			created_at, updated_at, visibility_state, guid, current_video_hls_url)
 		VALUES ($1, $2, $3, $4, $5,NOW(), NOW(), $6, $7)
```
joelhooks, line 86: we can write the media url on create as well

apps/egghead/src/lib/posts-query.ts
```diff
@@ -413,12 +416,25 @@ export async function writePostUpdateToDatabase(input: {
...
+
 	if (currentPost.fields.eggheadLessonId) {
 		await updateEggheadLesson({
 			eggheadLessonId: currentPost.fields.eggheadLessonId,
 			state: lessonState,
 			visibilityState: lessonVisibilityState,
 			duration: duration > 0 ? duration : timeToRead,
+			...(videoResource?.muxPlaybackId && {
+				hlsUrl: `https://stream.mux.com/${videoResource.muxPlaybackId}.m3u8`,
+			}),
```
joelhooks, line 437: basically any time we are writing the lesson we sync it to egghead db

PR #350, opened 2024-12-17
feat(egh): Course Creation Sync Flow

zacjones93:
Implements the course creation workflow that syncs courses between builder, egghead-rails, and sanity:

- Adds `egghead/course-created` event and handler
- Creates new course in egghead-rails when a course post is created
- Updates Sanity course creation to wait for Egghead course ID
- Adds TypeScript schemas for egghead-rails database courses and instructors

gif

Co-authored-by: Creeland <cree@provinsal.com>


apps/egghead/src/inngest/functions/egghead/create-course-in-egghead.ts
```diff
@@ -0,0 +1,87 @@
...
+		event: POST_CREATED_EVENT,
+	},
+	async ({ event, step }) => {
+		const post = await step.run('verify-post', async () => {
+			const { post } = event.data
+			if (!post) {
+				throw new NonRetriableError('Post not found')
+			}
+
+			if (post.fields?.postType !== 'course') {
```
joelhooks, line 26: this could be part of the trigger instead of here which imo makes more sense than throwing errors when any other kind of post is created

apps/egghead/src/inngest/functions/egghead/create-course-in-egghead.ts
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			return post
+		})
+
+		const instructor = await step.run('get-instructor', async () => {
+			const instructor = await loadEggheadInstructorForUser(post.createdById)
+
+			if (!instructor) {
+				return null
```
joelhooks, line 37: this could throw the error instead of returning null and probably remove the need for the guard below

apps/egghead/src/inngest/functions/egghead/create-course-in-egghead.ts
```diff
@@ -0,0 +1,87 @@
...
+			const course = await createEggheadCourse({
+				title: post.fields?.title,
+				guid: courseGuid,
+				ownerId: instructor.user_id,
+			})
+
+			return course
+		})
+
+		step.sendEvent(EGGHEAD_COURSE_CREATED_EVENT, {
```
joelhooks, line 64: kind of weird to go through all of the above and then send this different event 🤔

apps/egghead/src/inngest/functions/egghead/create-course-in-egghead.ts
```diff
@@ -0,0 +1,87 @@
...
+		if (!instructor) {
+			throw new NonRetriableError('Instructor not found', {
+				cause: {
+					postId: post.id,
+					instructorId: post.createdById,
+				},
+			})
+		}
+
+		const course = await step.run('create-course', async () => {
```
joelhooks, line 52: insider. among this eggheadCourse instead of just course because so many courses

apps/egghead/src/inngest/functions/sanity/create-course-in-sanity.ts
```diff
@@ -39,22 +40,38 @@ export const createCourseInSanity = inngest.createFunction(
 			return contributor
 		})
 
+		const courseCreatedEvent = await step.waitForEvent(
```
joelhooks, line 43: i'm not super stoked about the way these are connected together and think we might consider tightening it up or making the flow more obvious

this is the danger of event driven

entropy

apps/egghead/src/lib/users.ts
```diff
@@ -40,6 +40,36 @@ export const getCachedEggheadInstructorForUser = unstable_cache(
 	{ revalidate: 3600, tags: ['users'] },
 )
 
+export const InstructorDbSchema = z.object({
```
joelhooks, line 43: is the db here egghead? let's label it if so

joelhooks (approved):
i approved but there are a lot of suggestions for a follow up

PR #352, opened 2024-12-18
feat(ai): rss feed

vojtaholik:
image

gif


apps/ai-hero/src/app/rss.xml/route.ts
```diff
@@ -0,0 +1,50 @@
...
+		})
+
+	return feed.rss2()
+}
+
+export async function GET() {
+	return new Response(await generateRSS(), {
+		headers: {
+			'Content-Type': 'text/xml; charset=utf-8',
+			'Cache-Control': 's-maxage=1, stale-while-revalidate',
```
vojtaholik, line 47: @joelhooks 1 second intervals seem aggressive. I just went with what's on TT.
```suggestion
			'Cache-Control': 's-maxage=3600, stale-while-revalidate',
```

joelhooks: 1 is actually a very reasonable number for this and gives each user a cached experience while keeping it very close to always fresh

joelhooks (approved):

PR #355, opened 2024-12-19
fix: build

vojtaholik:
turns out that having typesense env vars defined in `.env.mjs` makes typesense client panic resulting in cryptic errors + ONLY during vercel build. this was hard to debug 😅 

gif


joelhooks:
"defined" or imported from? having them checked in the env module doesn't mean you need to import from there and still does the work of validating that they exist.

PR #357, opened 2024-12-20
feat(egh): add primary tag to post

zacjones93:
This adds a primary tag to posts.

`primaryTagId` is stored in `post.fields` and when a primary tag id is present we'll save the tag object to typesense so instructors have control over what tag is set as primary.

When a tag is selected for the first time it will automatically be selected as the primary tag. when there are multiple tags, users will be able to select the one they want as primary

on egghead we are using `image` which is just the first tag that happens to be tagged. This works fine in most cases but isn't flexible

typesense object

select for primary tag

gif


apps/egghead/src/lib/posts-query.ts
```diff
@@ -318,6 +318,45 @@ export async function addTagToPost(postId: string, tagId: string) {
...
+			.update(contentResource)
+			.set({
+				fields: {
+					...post.fields,
+					primaryTagId: tagId,
+				},
+			})
+			.where(eq(contentResource.id, postId))
+	}
+}
```
zacjones93, line 340: when we tag a post for the first time we want to automatically set that tag to the primary tag

apps/egghead/src/lib/typesense-query.ts
```diff
@@ -48,10 +46,12 @@ export async function upsertPostToTypeSense(post: Post, action: PostAction) {
 		})
 
 		const tags = await getPostTags(post.id)
+		const primaryTag = tags.find((tag) => post?.fields?.primaryTagId === tag.id)
+		const postGuid = post.id.split('_').pop()
 
 		const resource = TypesensePostSchema.safeParse({
 			id: `${post.fields.eggheadLessonId}`,
-			externalId: post.id,
+			externalId: postGuid,
```
zacjones93, line 54: `externald` should be the guid where id of post is `post_:guid`

apps/egghead/src/lib/typesense-query.ts
```diff
@@ -60,6 +60,12 @@ export async function upsertPostToTypeSense(post: Post, action: PostAction) {
 			path: `/${post.fields.slug}`,
 			type: post.fields.postType,
 			_tags: tags.map((tag) => tag.fields.name),
+			...(primaryTag && {
+				primary_tag: primaryTag,
+				...(primaryTag.fields?.image_url && {
+					primary_tag_image_url: primaryTag.fields.image_url,
+				}),
+			}),
```
zacjones93, line 68: conditionally set primary tag and primary_tag_image_url when it's present

packages/ui/resources-crud/tag-selector-with-primary.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+type Tag = {
+	id: string
+	fields: {
+		label: string
+		name: string
+	}
+}
+
+export default function AdvancedTagSelectorWithPrimary({
```
zacjones93, line 30: added `AdvancedTagSelectorWithPrimary` as a new component in ui package instead of modifying the existing one

packages/ui/resources-crud/tag-selector-with-primary.tsx
```diff
@@ -0,0 +1,196 @@
...
+	onPrimaryTagSelect?: (tagId: Tag['id']) => void
+	primaryTagId: Tag['id'] | null
+}) {
+	const [selectedPrimaryTag, setSelectedPrimaryTag] = React.useState<Tag['id']>(
+		primaryTagId ?? '',
+	)
+
+	React.useEffect(() => {
+		setSelectedPrimaryTag(primaryTagId ?? '')
+	}, [primaryTagId])
```
zacjones93, line 158: when a tag is selected for the first time, we automatically set the primary tag id for the post. we need to pick up on that change with this useEffect

apps/egghead/src/app/(content)/posts/_components/edit-post-form-metadata.tsx
```diff
@@ -237,15 +241,23 @@ export const PostMetadataFormFields: React.FC<{
...
+				<FormDescription>
+					The selected tag will be used as the primary tag for the post.
+				</FormDescription>
+				<AdvancedTagSelectorWithPrimary
 					availableTags={tags}
 					selectedTags={post?.tags?.map((tag) => tag.tag) ?? []}
+					primaryTagId={post?.fields?.primaryTagId ?? ''}
 					onTagSelect={async (tag: { id: string }) => {
 						await addTagToPost(post.id, tag.id)
+						router.refresh()
```
zacjones93, line 253: `router.refresh` to pick up on `primaryTagId` changing to display immediately to user

packages/ui/resources-crud/tag-selector-with-primary.tsx
joelhooks: personally I think that "canSelectPrimary" or similar might make sense on the single tag selector versus making another one

something to keep considering 🤔

joelhooks (approved):

PR #359, opened 2024-12-23
feat: playlist syncing backtrack

joelhooks:


apps/egghead/src/app/api/posts/route.ts
```diff
@@ -93,7 +93,9 @@ export async function POST(request: NextRequest) {
 
 	try {
 		const newPost = await writeNewPostToDatabase({
-			newPost: validatedData.data,
+			title: validatedData.data.title,
+			videoResourceId: validatedData.data.videoResourceId || undefined,
+			postType: validatedData.data.postType,
```
joelhooks, line 98: instead of passing the full objects we just pass the data we need

apps/egghead/src/inngest/events/post-created.ts
```diff
@@ -4,6 +4,6 @@ export const POST_CREATED_EVENT = 'post/created'
 export type PostCreated = {
 	name: typeof POST_CREATED_EVENT
 	data: {
-		post: ContentResource | null
+		post: ContentResource
```
joelhooks, line 7: imo this should ALWAYS have a value or not get sent

apps/egghead/src/lib/egghead.ts
```diff
@@ -76,6 +76,8 @@ export async function getEggheadLesson(eggheadLessonId: number) {
 	return lesson
 }
 
+// createEggheadCourse
+
```
joelhooks, line 80:
```suggestion
```

apps/egghead/src/lib/posts-query.ts
```diff
@@ -260,7 +259,7 @@ export async function createPost(input: NewPost) {
 		await inngest.send({
 			name: POST_CREATED_EVENT,
 			data: {
-				post: await post,
+				post: post,
 			},
 		})
```
joelhooks, line 265: @zacjones93 i didn't get why the `await` was on the post instead of above. the ramification is that the if statement would always be true, since `writeNewPostToDatabase` returns a promise 🤔

apps/egghead/src/lib/posts-query.ts
```diff
@@ -349,106 +348,6 @@ export async function removeTagFromPost(postId: string, tagId: string) {
 	await writeLegacyTaggingsToEgghead(postId)
 }
 
-export async function writeNewPostToDatabase(input: {
```
joelhooks, line 352: this file is so huge thought it'd make sense to yoink this out into it's own file

apps/egghead/src/lib/posts-query.ts
```diff
@@ -590,60 +489,6 @@ export async function deletePostFromDatabase(id: string) {
 	return true
 }
 
-export async function createNewPostVersion(
```
joelhooks, line 593: same here, sent to it's own file

apps/egghead/src/lib/sanity-content-query.ts
```diff
@@ -208,6 +208,9 @@ export async function getSanityLessonForEggheadLessonId(
 	)
 }
 
+// createSanityCourse
```
joelhooks, line 211:
```suggestion
```

apps/egghead/src/lib/errors/post-errors.ts
```diff
@@ -0,0 +1,33 @@
...
+export class DatabaseError extends PostCreationError {
+	constructor(
+		operation: string,
+		cause?: unknown,
+		context?: Record<string, unknown>,
+	) {
+		super(`Database operation failed: ${operation}`, cause, context)
+		this.name = 'DatabaseError'
+	}
+}
```
joelhooks, line 33: this was generated to add better error handling and seemed like a nice idea

PR #383, opened 2025-01-20
refactor(glf): refactor for loop to flatMap with Promise.all

zacjones93:
refactors lesson list `saveTreeData` to use `Promise.all` in a queue. 

gif


vojtaholik (commented):
I think this should be added to ai hero too and tested

vojtaholik (approved):

PR #397, opened 2025-01-27
feat(egh): remove post from course post

zacjones93:
This follows the same pattern used to remove posts as implemented for [adding posts](https://github.com/badass-courses/course-builder/commit/e19baf792a6b27536b434ca336e5fcf6226765e7).

Another PR should be considered to consolidate this syncing logic into it's own lib file

gif


apps/egghead/src/app/(content)/posts/_components/resource-resources-list.tsx
```diff
@@ -127,6 +130,23 @@ export function ResourceResourcesList({
...
+					await removePostFromCoursePost({
+						postId: itemId,
+						resourceOfId: resource.id,
+					})
+
+					await removeEggheadLessonFromPlaylist({
+						eggheadLessonId: post.fields?.eggheadLessonId,
+						eggheadPlaylistId: resource.fields?.eggheadPlaylistId,
+					})
+				}}
```
zacjones93, line 149: following pattern seen in `EditResourcesForm` where we define logic at the top level and pass function down to where it is called

apps/egghead/src/app/(content)/posts/actions.ts
```diff
@@ -0,0 +1,22 @@
...
+	resourceOfId: string
+}) {
+	return await db
+		.delete(contentResourceResource)
+		.where(
+			and(
+				eq(contentResourceResource.resourceOfId, resourceOfId),
+				eq(contentResourceResource.resourceId, postId),
+			),
+		)
```
zacjones93, line 21: needs to run on the server so put function in actions folder. Reviewing it now, this should probably be in `posts-query`?

apps/egghead/src/components/lesson-list/pieces/tree/tree-item.tsx
```diff
@@ -404,12 +406,12 @@ const TreeItem = memo(function TreeItem({
 						</span>
 					) : null} */}
 				</button>
-				{item.type === 'section' && item.children.length === 0 && (
+				{item.type === 'post' && (
```
zacjones93, line 409: we don't have sections in egghead right now so switched this block for post deletion

apps/egghead/src/lib/posts-query.ts
```diff
@@ -624,6 +624,52 @@ export async function addEggheadLessonToPlaylist({
...
+					},
+				}),
+			},
+		).then(async (res) => {
+			if (!res.ok) {
+				console.log('Error removing lesson from playlist', res)
+				throw new EggheadApiError(res.statusText, res.status)
+			}
+			return await res.json()
+		})
```
zacjones93, line 665: the syncing work

zacjones93: @Creeland and I went this initial route because of the tracklist table/logic involved with playlists but we can look at refactoring these api calls to DB operations as a next step

joelhooks: ya it's definitely simpler/safer but more coupled to rails so tradeoff is a thing. It's fine, just something to consider. Wouldn't change for this PR for sure.

apps/egghead/src/app/(content)/posts/_components/resource-resources-list.tsx
```diff
@@ -127,6 +130,21 @@ export function ResourceResourcesList({
...
+							resourceOfId: resource.id,
+						})
+					} catch (error) {
+						console.error('Error removing lesson from playlist', error)
+						toast({
+							title: 'Error removing lesson from playlist',
+							description: 'Please refresh the page and try again.',
+							variant: 'destructive',
+						})
+					}
```
zacjones93, line 146: https://www.loom.com/share/3dd17151bdd94a559db99770466c2293?sid=a6ecd7b2-df5f-404d-9bf4-e86e4d18c7f6

Added some error handling that lets user know something went wrong in a toast

Creeland (approved):

PR #406, opened 2025-02-12
feat(egh): publish egghead playlist onPublish

zacjones93:
gif


apps/egghead/src/lib/egghead.ts
```diff
@@ -470,6 +473,16 @@ export async function createEggheadCourse(input: {
...
+export function setPlaylistPublishedAt(
+	eggheadPlaylistId: number,
+	publishedAt: string,
+) {
+	return eggheadPgQuery(
+		`UPDATE playlists SET published_at = $1 WHERE id = $2`,
+		[publishedAt, eggheadPlaylistId],
+	)
+}
+
```
zacjones93, line 485: is it better to have one function and switch on the type? First instinct is to avoid dynamic sql

joelhooks: I don't fully understand the question yet. Looks fine to me.

joelhooks: i think you could make both of these updates in a single query

joelhooks (approved):

PR #409, opened 2025-02-14
feat: cohort content resources with tier metadata

joelhooks:
we need to be able to add content to specific cohorts and designate the content for specific tiers standard premium VIP

https://www.loom.com/share/9d64401e9ca34bd296085ac7d92c409f


vojtaholik (approved):
🙇‍♂️

PR #418, opened 2025-02-19
feat(egh): add course info to TypeSense

zacjones93:
The Feed on egghead-next gets clogged up with course lessons if an instructor publishes them all at once. This adds `resources` array and `belongs_to_resource` boolean to typesense data so that we can sort (effectively filter out) lessons in courses in a new preset called `the_feed`. 

TypeSense does not let you filter on null values but the workaround is to sort missing values to the bottom (`belongs_to_resource(missing_values: first):desc`). We set `belongs_to_resource` to the id of the resource the current resource belongs to.

We also weren't updating the egghead playlist on publish to properly index it so this adds that as well

related
https://github.com/skillrecordings/egghead-next/pull/1508

gif

### DIFF
--- a/apps/egghead/src/lib/typesense.ts
+++ b/apps/egghead/src/lib/typesense.ts
@@ -29,8 +29,8 @@ export const TypesensePostSchema = z.object({
 	instructor_url: z.string().url().optional(),
 	path: z.string().optional(),
 	published_at_timestamp: z.number().nullish(),
-	belongs_to_course: z.number().nullish(),
-	courses: z
+	belongs_to_resource: z.number().nullish(),
+	resources: z
 		.array(
 			z.object({
 				id: z.string(),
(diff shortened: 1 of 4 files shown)


apps/egghead/src/lib/typesense.ts
```diff
@@ -29,6 +29,17 @@ export const TypesensePostSchema = z.object({
 	instructor_url: z.string().url().optional(),
 	path: z.string().optional(),
 	published_at_timestamp: z.number().nullish(),
+	belongs_to_course: z.boolean().optional(),
```
joelhooks, line 32: i think instead of a boolean this should be the Id of the playlist 

also this is very specific "course" but might also include other list types? 🤔

zacjones93: updated to `belongs_to_resource` and `resources`

PR #480, opened 2025-05-27
feat(aih): implement cohort transfer workflow for purchase management

vojtaholik:
gif

### DIFF
--- a/apps/ai-hero/src/app/(commerce)/subscribe/verify-login/page.tsx
+++ b/apps/ai-hero/src/app/(commerce)/subscribe/verify-login/page.tsx
@@ -69,10 +69,10 @@ export default async function VerifyLoginPage({
 	return (
 		<LayoutClient withContainer>
 			<Login
-				title="Login to Subscribe"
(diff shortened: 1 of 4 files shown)


apps/ai-hero/src/server/auth.ts
```diff
@@ -229,35 +230,15 @@ export const authOptions: NextAuthConfig = {
...
 
-			const currentMembership = organizationId
-				? await db.query.organizationMemberships.findFirst({
-						where: and(
-							eq(organizationMemberships.organizationId, organizationId),
-							eq(organizationMemberships.userId, user.id),
-						),
-					})
-				: null
+			// Load entitlements from ALL user organizations, not just the current one
```
joelhooks, line 238: from a hygiene perspective you wouldn't load all of their entitlements and they should switch orgs to gain access to other org entitlements


in practice it's probably fine, but that's the idea

joelhooks: coderabbit feels the same below i noticed

